PR #10, opened 2019-05-22
logger.exception & disable logging.config

keisuke-nakata:
logging is preferred to be configured at user side (in order to let user configure their own settings)


keisuke-nakata:
Why this PR was closed?

MadcowD:
We have integrated these changes already :)

PR #12, opened 2019-06-05
suppress error `AttributeError: 'MineRLEnv' object has no attribute '_log_error'`

keisuke-nakata:


keisuke-nakata:
Current master does not have this issue

PR #15, opened 2019-06-08
Fixing README sample code (Random agent)

nicomon24:
Fixing README sample code with sample() instead of random() for the action space


MadcowD:
Thanks so much

PR #25, opened 2019-06-10
Fix bug where existing data_dir passed as argument was not detected

albertwujj:


MadcowD:
Looks good! Thanks for catching this.

Issue #30, opened 2019-06-10
Documentation Recommends Use of Deprecated Brew Java8 Installation


decodyng:
This is more a FYI about documentation than it is a bug in code, but trying to run the commands that you recommend in the docs to install Java8 on a mac, namely: 

```
brew tap caskroom/versions
brew cask install java8 
```

errors out with 

```
Error: Cask 'java8' is unavailable: No Cask with this name exists.
```

Per this stack overflow, it appears that installing java8 via brew is now best done through AdoptOpenJDK: https://stackoverflow.com/questions/55834845/unable-to-install-java8-with-homebrew

MadcowD:
Oh awesome find! Would you be willing to make a PR for this in the documentation?

decodyng:
Sure, I'd be happy to

MadcowD:
:)))))

decodyng:
PR here: https://github.com/minerllabs/minerl/pull/32

Apologies for it containing about as many commits as there are words in the sentence I'm changing; I'm new to RST and got a bit burned by how fiddly it is.

MadcowD:
I just learned how to do it last weekend . Believe me I know :/

minerllabs:
Closed by #32

PR #32, opened 2019-06-10
Fix Documentation to Include Working Mac Java8 Installation Instructions

decodyng:
Per #30 , this PR contains modifications of the docs to use AdoptOpenJDK brew cask to install Java8, to deal with the fact that Oracle has deprecated the original links for the java8 binaries used in the old instructions. Also includes a few small spelling fixes.


MadcowD:
Thanks so much!

PR #35, opened 2019-06-10
Overriding env done flag when info dict is empty

### LINKED ISSUE #24: Observation returned by env.step() changes key from 'inventory' to 'dirt' near navigate goal
I was following along the tutorial in the docs and making some of my own edits to the base code and found that in the two time steps right before done changes to True, the 'inventory' key is replaced with a 'dirt' key, with the value also becoming an `OrderedDict`. The structure of the returned observation looks fine otherwise.

On a tangent, could you add bullet pointing for the documentation of `env.step()`? Currently it's one block of text and looks incomplete.

nicomon24:
Fixes #24:
It seems there is a lag of one timestep between the real end of the environment and the done flag.
In particular, by inspecting the info dict from _core.py_, line 483:
`info = comms.recv_message(self.client_socket).decode('utf-8')`
in a Navigate environment, we can see that, at the second-last timestep, the info dict contains 
`"WorldTime":6000`, 
which means the episode is effectively ended in the Malmo simulator (by mission definition).
In the following step, we see not only a done flag which is still False, but also an empty info dict (because the simulator has terminated the mission). 
This causes the environment to step one more time, resulting in the strange behaviour of #24.

I propose this hot-fix where we directly set the done flag to True if the info dict is empty. While this solves the issue, it would be better to further inspect this interaction between Malmo and mineRL to avoid further unnoticed bugs.


MadcowD:
Thanks so much for bringing this to our attention! I'm checking through the environment to see if we can make this change on the Malmo side today,

nicomon24:
Just realized this fix does also remove the last reward, so if the agent reaches the diamond block (in Navigate), it doesn't get the +100 reward at all. I'm closing the PR since this is not perfectly working, maybe I'll look into the Malmo environment to see if we can get it to work as it is supposed to be!

MadcowD:
I believe this is fixed with `minerl` > 0.1.10. Please upgrade to see if this fixes your issue
```
pip3 install --upgrade minerl
```

PR #61, opened 2019-06-18
Actually generate new Worlds for Treechop

TimZF:
Change so a new World gets generated instead of reusing the same one


MadcowD:
Hey taking a look at this! 

We were hoping that `destroyAfterUse` would actually do this.

PR #98, opened 2019-06-25
Fixing the broke pipe error, and a bunch of other bugs.

MadcowD:
Several fixes around the central infrastructure.


MadcowD:
Closes #69

PR #106, opened 2019-06-27
Add seed parameter to seed() method

lbertge:
In class `MineRLEnv`, the `seed()` method doesn't properly extend its parent method. The `gym.Env` parent class has the method signature `seed(self, seed=None)`, so I think the `MineRLEnv` class should do the same. Sorry about all the other changed lines, my formatter clears any unnecessary whitespace. The important line is [here](https://github.com/minerllabs/minerl/compare/master...lbertge:seed-signature?expand=1#diff-bd16a01d42aeeabefbf0d18b10df8f65R477).

As an aside, I'd like to know why seeds aren't implemented. It seems like it'd be important to have for reproducibility. If it's not too much trouble I could look into working on it.

### DIFF
--- a/minerl/env/core.py
+++ b/minerl/env/core.py
@@ -49,7 +49,7 @@ class EnvException(Exception):
     """A special exception thrown in the creation of an environment's Malmo mission XML.
 
     Args:
-        message (str): The exception message. 
+        message (str): The exception message.
     """
 
     def __init__(self, message):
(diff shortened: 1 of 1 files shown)


MadcowD:
Wow thanks for this fix!
It would be fantastic if you could look into seeds. The main gist of this is that the we need to seed every instance of the Minecraft Random number generator. Perhaps one could write a mixin to replace the java.util.Random class with one which supports global seeding.

PR #107, opened 2019-06-27
Brandon/unequip

### LINKED ISSUE #102: equip unequip
For the craft actions, `none` is fine, but for equip, they may unequip the current item, for which `equip` `unequip` seems like an appropriate action to be added to the list.

brandonhoughton:
Adds ability to un-equip via `equip air` command
Adds observation of empty mainhand 
Adds documentation for equip action

closes #102


brandonhoughton:
@MadcowD Requires re-publishing of documentation

PR #111, opened 2019-06-28
Add simple test, move broken tests to excluded dir

brandonhoughton:
We can integrate env and data tests when those elements are accounted for in the travis config


MadcowD:
Hey your simple test should close the envs after every env.make

MadcowD:
At least try resetting them as well

brandonhoughton:
Good to know that envs need to be closed. For now pending the headless fix for the travis ci build system I will stick to simply importing the envs through the package and using gym.spec to ensure they were registered properly

PR #121, opened 2019-07-03
remove always empty dict

lbertge:
A successful `_peek_obs` call always returns a tuple containing the observation and an empty dictionary. This seems unnecessary, since all other methods on the function stack just seem to return the result as is, without modification.


MadcowD:
I agree, my only issue is that this might invalidate people's existing code. Do you have any thoughts on how to make this transition easier?

lbertge:
Oh, I didn't properly think about the consequences. Right now there are two ways that I see:
1. Log a warning in the method about the API change, and then merge after some versions (my estimate is that 1 month is enough time for that). The issue with this is that people won't be able to use the updated API until whenever this branch is merged.
2. Copy into a second `_peek_obs_v2` method, and give the user the ability to set a flag which directs the code to use the old method, or the new method. This is smoother for transitioning but more clunky, since this distinction should be abstracted from users.

Personally I think that supporting a transition is a hassle and not worth the effort for cleaning up the API. I guess my recommendation would be to bite the bullet and force users to change, or just leave as is (and close this PR). I'm OK with either one, though I'm sure other devs would prefer the latter.

MadcowD:
Okay! Thanks for the proposal, we'll chat internally. We luckily have the benefit of being in alpha so such a change shouldn't be too much of an issue.

I'll get back to you soon.

MadcowD:
Hey! Our team has discussed this, and we've decided to keep this feature as we plan on eventually using this dictionary (the info dict on all observations) to send auxiliary information. Thank you so much for being thoughtful and proposing this solution!

:) 
Will

lbertge:
This PR can stay closed, but I respectfully ask to have another glance at the issue. This PR is about changes to `_peek_obs()`, which is called whenever the environment is reset. That means I'll have an API like:
`obs, info = env.reset()`

But if I then step into the environment, the API will look like:
```
obs, _, _, _ = env.step()
assert isinstance(obs, dict)
```

This is because under the hood `step()` calls on `_process_observation()`, which returns a single dict containing all the observations. So I won't have access to auxiliary information while stepping.

If this is truly intended, then I'll concede here. But my guess it's not, and what you really wanted to do (as [gym's documentation](http://gym.openai.com/docs/) suggests), is to include the `info` from the step, and not the reset:
```
obs = env.reset()
obs, reward, done, info = env.step()
```
which means you'll have to find some way to get `info` into step. Probably the easiest way to do that is to modify the `_process_observation` method, but I won't get into it here. 

P.S. If I have a wrapper on the env, I'm forced to do something like this:
```
def observation(self, observation):
  if isinstance(observation, tuple):
          observation = observation[0]
  # do something with observation
```
which would not break with the API change, since we have to check anyways if the incoming observation is a tuple or not.

PR #149, opened 2019-07-16
allow base port for malmo to be configured

AndKram:


brandonhoughton:
Looks good to me! Issue with Travis CI seems unrelated - re-running the test after I fix the issue with pySmartDL casting None to int

Issue #198, opened 2019-08-04
MineRLEnv spec.id assumes that class is instantiated through gym


tomsilver:
[This line](https://github.com/minerllabs/minerl/blob/master/minerl/env/core.py#L208) in `core.py` references `self.spec.id`, which exists only if `MineRLEnv` is instantiated through`gym` (c.f. [gym registration]( https://github.com/openai/gym/blob/master/gym/envs/registration.py#L95).) The class is otherwise supportive of `MineRLEnv` instances being instantiated directly, so it seems like low hanging fruit to change this line. I will open a PR (my first!) shortly with a possible fix.

PR #199, opened 2019-08-04
Remove MineRLEnv dependency on self.spec.id

tomsilver:
Addresses https://github.com/minerllabs/minerl/issues/198 . I'm not sure if 'Custom MineRL Env' is the best default. Thanks!

### DIFF
--- a/minerl/env/core.py
+++ b/minerl/env/core.py
@@ -71,7 +71,7 @@ class MissionInitException(Exception):
 
 MAX_WAIT = 80  # After this many MALMO_BUSY's a timeout exception will be thrown
 SOCKTIME = 60.0 * 4  # After this much time a socket exception will be thrown.
-
+MINERL_CUSTOM_ENV_ID = 'MineRLCustomEnv' # Default id for a MineRLEnv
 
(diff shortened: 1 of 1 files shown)


minerl/env/core.py
```diff
@@ -205,7 +205,12 @@ def init(self):
             xml = f.read()
         # Todo: This will fail when using a remote instance manager.
         xml = xml.replace('$(MISSIONS_DIR)', missions_dir)
-        xml = xml.replace('$(ENV_NAME)', self.spec.id)
+
+        if self.spec is not None:
+            xml = xml.replace('$(ENV_NAME)', self.spec.id)
+        else:
+            xml = xml.replace('$(ENV_NAME)', 'Custom MineRL Env')
```
MadcowD: Looks fine, if you could change the string to 'MineRLCustomEnv'  and move it to a global constant (
`MINERL_CUSTOM_ENV_ID='MineRLCustomEnv'`) at the top of the file, then this is good to go!

tomsilver:
@MadcowD good call, thanks! Let me know if it looks good now.

PR #300, opened 2020-06-08
update matplotlib requirements

keisuke-nakata:
On macOS and Python 3.8.2, matplotlib==3.0.3 fails to build/install.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,4 +11,4 @@ psutil>=5.6.2
 Pyro4>=4.76
 getch>=1.0; sys_platform != 'win32' and sys_platform != 'cygwin'
 coloredlogs>=10.0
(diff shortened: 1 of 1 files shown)


brandonhoughton:
@MadcowD This version of Matplot lib was exact for a reason. Do you know what breaks so I can test newer versions?

keisuke-nakata:
`dill` does not work on Python3.8.2, so I don't mind Python 3.8 anymore.
I understand there is a reason for fixing the version of matplotlib and I think upgrading is not high priority, so I will close this PR.
Thank you!

ntoxeg:
Is the above problem with `dill` still persisting? I could not find any open issues about Python 3.8 incompatibility anymore on their GitHub.

PR #308, opened 2020-06-14
Randomized port allocation with retry

shwang:
Fixes a race condition where MineRLEnv fails to launch if multiple MineRLEnv are launched in parallel from different process (e.g. 5 different seeds of the same training script).

On my machine, using this branch, when I launch 5 MineRLEnv in parallel,
there is still a ~20% chance of MineRLEnv failing due to race condition
that reports corrupted Gradle pack error. (Fixed later by HumanCompatibleAI/pull/6).

## Original PR description from HumanCompatibleAI/pull/5:

We run into a race condition when multiple processes launch MineRL environments from different InstanceManagers, and different Minecraft instances are instructed to bind to the same port.

While connecting to Minecraft, MineRLEnv now looks for `BindException` inside the Minecraft process's logs (indicating a port collision) and immediately relaunch a new Minecraft instance with a different port when this happens. This allows us to recover from the race condition, but is expensive because launching a Minecraft instance takes time.

The InstanceManager port allocation scheme now chooses randomly from unbound ephemeral ports instead of linearly searching through (9000, 9001, ...). This reduces the chance of a race condition arising.

Other changes:
* Removed `InstanceManager._malmo_base_port`. This used to be used primarily as a starting point for linearly searching for available ports. Since we randomly allocate now this isn't necessary.
* Removed `InstanceManager.is_display_port()`, as it always returned False and depended on `InstanceManager._malmo_base_port`.

### DIFF
--- a/minerl/env/core.py
+++ b/minerl/env/core.py
@@ -703,12 +703,40 @@ class MineRLEnv(gym.Env):
                 num_retries += 1
                 if num_retries > MAX_WAIT:
                     raise socket.timeout()
-                logger.debug("Recieved a MALMOBUSY from Malmo; trying again.")
-                time.sleep(1)
-
+                elif self.log_shows_bind_exception():
+                    # Skips error handler to directly to abort.
+                    self.had_to_clean = True
+                    logger.error("Malmo server failed to bind to port, possibly due "
+                                 "to collision with parallel Malmo instance. "
+                                 "Giving up on contacting this Malmo server, and "
+                                 "starting a new one. It's possible that the abandoned Malmo "
+                                 "server cannot be closed automatically, and you will "
+                                 "have to do so manually later.")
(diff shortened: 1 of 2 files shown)


shwang:
Fails on py3.5 build because the patch added an f-string. I'll remove that.

The other builds are failing on `minerl.data.download`.

brandonhoughton:
Yeah we were debating python 3.5 support because f-strings are so clean, but it doesn't make sense to drop from a usability standpoint.

shwang:
This PR is no longer necessary after https://github.com/minerllabs/minerl/pull/260, though I argue this is a simpler solution to the same problem.

Are the malmo base port and display port variables necessary for the project in the current form? If they aren't used by anything then they just add complexity to the port assignment logic.

PR #317, opened 2020-06-17
Update .travis.yml

brandonhoughton:


brandonhoughton:
No longer needed because we have moved to circleCI

PR #320, opened 2020-06-17
Fix VectorObf envs timing out (issue #319)

### LINKED ISSUE #319: *VectorObf-v0  envs do not trigger timeout
Running the example code below does not trigger episode timeout, which I assume [should be at 8000 steps](https://github.com/minerllabs/minerl/blob/ce8731f8ea43da211c27e6f4d34d07d6c7aa9007/minerl/herobraine/env_specs/treechop_specs.py#L35). It seems to continue forever / far too long, or until player dies. Tested with TreeChop and ObtainDiamond envs with VectorObf variants.

I am not sure if same happens with original environments. I am trying to test this but agent keeps dying to night-mobs :). Dang creepers bombing development, too.

**Edit:** Hmm, seems like `env.env_spec.max_episode_steps` is `None` for Obf envs. For non-obf environments it seems correct (e.g. 8000 for TreeChop).

```python3
import gym
import minerl

if __name__ == "__main__":
    # env = gym.make("MineRLObtainDiamondVectorObf-v0")
    env = gym.make("MineRLTreechopVectorObf-v0")
    obs = env.reset()
    done = False
    step_ctr = 0
    while not done:
        _ = env.step(env.action_space.sample())
        step_ctr += 1
        if (step_ctr % 1000) == 0:
            print("Current step: {}".format(step_ctr))
```

System/versions:
* Happens on Windows 10 and Ubuntu 16.04
* Python 3.6
* MineRL 0.3.1 (pypi)

Miffyli:
Fixes #319 

Updates EnvWrapper to use environment's `max_episode_steps` and `reward_threshold` (like `xml` was passed), rather than forcing them to `None`

### DIFF
--- a/minerl/herobraine/wrapper.py
+++ b/minerl/herobraine/wrapper.py
@@ -17,8 +17,9 @@ class EnvWrapper(EnvSpec):
             self._unwrap_act_fn = self.env_to_wrap.unwrap_action
             self._unwrap_obs_fn = self.env_to_wrap.unwrap_observation
 
-        super().__init__(self._update_name(env_to_wrap.name), env_to_wrap.xml, max_episode_steps=None,
-                         reward_threshold=None)
+        super().__init__(self._update_name(env_to_wrap.name), env_to_wrap.xml,
+                         max_episode_steps=env_to_wrap.max_episode_steps,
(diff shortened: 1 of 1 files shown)


brandonhoughton:
Awesome! Thanks for this

Issue #210, opened 2019-08-10
Cannot start multiple environments on windows (shutil permissions issue)


katja-hofmann:
An exception is thrown when trying to create a second minerl environment instance on Windows. Tested with minerl 0.2.0 and 0.2.3 

```
> python
Python 3.6.6 (v3.6.6:4cf1f54eb7, Jun 27 2018, 03:37:03) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import gym
>>> import minerl
>>> env = gym.make('MineRLTreechop-v0')
>>>
>>> import logging
>>> logging.basicConfig(level=logging.DEBUG)
>>>
>>> env2 = gym.make('MineRLTreechop-v0')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\test\AppData\Roaming\Python\Python36\site-packages\gym\envs\registration.py", line 156, in make
    return registry.make(id, **kwargs)
  File "C:\Users\test\AppData\Roaming\Python\Python36\site-packages\gym\envs\registration.py", line 101, in make
    env = spec.make(**kwargs)
  File "C:\Users\test\AppData\Roaming\Python\Python36\site-packages\gym\envs\registration.py", line 73, in make
    env = cls(**_kwargs)
  File "C:\Program Files\Python36\lib\site-packages\minerl\env\core.py", line 133, in __init__
    self.instance = self._get_new_instance(port)
  File "C:\Program Files\Python36\lib\site-packages\minerl\env\core.py", line 155, in _get_new_instance
    instance.launch()
  File "C:\Program Files\Python36\lib\site-packages\minerl\env\malmo.py", line 438, in launch
    shutil.copytree(os.path.join(InstanceManager.MINECRAFT_DIR), self.minecraft_dir)
  File "C:\Program Files\Python36\lib\shutil.py", line 359, in copytree
    raise Error(errors)
shutil.Error: [('C:\\Program Files\\Python36\\lib\\site-packages\\minerl\\env\\Malmo\\Minecraft\\.gradle\\2.14\\taskArtifacts\\cache.properties.lock', 'C:\\Users\\test\\AppData\\Local\\Temp\\tmpc09k86if\\Minecraft\\.gradle\\2.14\\taskArtifacts\\cache.properties.lock', '[Errno 13] Permission denied')]
```

MadcowD:
[...]

PR #334, opened 2020-06-29
Multiple environment test

juliusfrost:
Adds a simple test with multiple environments. The test spawns two threads and tries to run the agent-environment loop. The goal is to pass this test without errors so minerl is thread-safe. It supports synchronous environment creation and asynchronous environment creation. Running this reproduces #210 on Windows


MadcowD:
Wow thank you!

PR #352, opened 2020-07-13
Thread-safe launching of minecraft instance to solve creating multiple minerl environments in parallel

juliusfrost:
Adds a simple thread lock to prevent multiple processes from accessing the launch script. This prevents gradle build errors when trying to create multiple minerl environments in parallel.

### DIFF
--- a/minerl/env/core.py
+++ b/minerl/env/core.py
@@ -28,6 +28,7 @@ import socket
 import struct
 import time
 import uuid
+import threading
 from copy import copy, deepcopy
 from typing import Iterable
 
@@ -46,6 +47,8 @@ from minerl.herobraine.wrapper import EnvWrapper
 logger = logging.getLogger(__name__)
 
 missions_dir = os.path.join(os.path.dirname(__file__), 'missions')
+
+launch_instance_lock = threading.Lock()
 
 
 class EnvException(Exception):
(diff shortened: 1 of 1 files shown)


minerllabs:
Does this require waiting for one instance to lock then launching the next?

juliusfrost:
@minerllabs Yes, it does. There could be a way to narrow the scope of the lock period, but so far as I see it, there isn't a good way as the logic is controlled in the Minecraft client. But in practice, I didn't see much of an issue with the locking.

juliusfrost:
I found that this solution doesn't work with certain multiprocessing techniques to parallelize the environments, such as in RLlib. So this pull might not be the best to merge yet. I'll be looking into this.

minerl/env/core.py
```diff
@@ -47,6 +48,8 @@
 
 missions_dir = os.path.join(os.path.dirname(__file__), 'missions')
 
+launch_instance_lock = FileLock(os.path.join(os.path.dirname(__file__), 'env.lock'))
```
juliusfrost, line 51: FileLock uses a file as communication for processes to proceed in the code. When a process enters the launch instance code, FileLock creates a file `env.lock` in the same directory and prevents other processes from entering unless the lock is removed.

juliusfrost:
@MadcowD @minerllabs What should be done further to merge this?

neevparikh:
Hey this would be nice to have @MadcowD @minerllabs. Can we approve this PR?

danijar:
Any update on this?

Miffyli:
This seems to have been fixed in the latest version. At least [this test script](https://github.com/minerllabs/minerl/blob/dev/tests/excluded/multiple_env_test.py) runs fine (but for some reason runs slow on my machine, even with a single environment. Might be my mangled environments). Parallel instances do not work on Windows, tho.

PR #409, opened 2020-09-23
first pass at making multiagent compatible with openai code

AdrienLE:
Note: needs to be merged at the same time as the corresponding branch in the monorepo

### DIFF
--- a/minerl/env/_multiagent.py
+++ b/minerl/env/_multiagent.py
@@ -551,6 +551,9 @@ class _MultiAgentEnv(gym.Env):
         self.instances = self.instances[:self.task.agent_count]
 
         # Now let's clean and establish new socket connections.
+        # Note: it is important that all clients are informed of the episode end BEFORE the
+        #  server. Since the first client is the one that communicates to the server, we
+        #  inform it last by iterating backwards.
         for instance in reversed(self.instances):
             self._TO_MOVE_clean_connection(instance)
             self._TO_MOVE_create_connection(instance)
--- a/minerl/env/_multiagent.py
+++ b/minerl/env/_multiagent.py
@@ -606,7 +606,8 @@ class _MultiAgentEnv(gym.Env):
             ok, = struct.unpack("!I", reply)
             if ok != 1:
                 num_retries += 1
-                # TODO: This is odd, MAX_WAIT is usually a number of seconds but here it's a number of retries. Probably needs to be fixed.
+                # TODO: This is odd, MAX_WAIT is usually a number of seconds but here
+                #  it's a number of retries. Probably needs to be fixed.
                 if num_retries > MAX_WAIT:
                     raise socket.timeout()
                 logger.debug("Recieved a MALMOBUSY from {}; trying again.".format(instance))
--- a/minerl/env/_multiagent.py
+++ b/minerl/env/_multiagent.py
@@ -551,7 +551,7 @@ class _MultiAgentEnv(gym.Env):
         self.instances = self.instances[:self.task.agent_count]
 
         # Now let's clean and establish new socket connections.
-        for instance in self.instances:
+        for instance in reversed(self.instances):
             self._TO_MOVE_clean_connection(instance)
             self._TO_MOVE_create_connection(instance)
             # The socket could be failed here. This method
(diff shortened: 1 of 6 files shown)


JoostHuizinga (approved):
Looks good to me.

minerl/env/_multiagent.py
```diff
@@ -551,7 +551,7 @@ def _setup_instances(self) -> None:
         self.instances = self.instances[:self.task.agent_count]
 
         # Now let's clean and establish new socket connections.
-        for instance in self.instances:
+        for instance in reversed(self.instances):
```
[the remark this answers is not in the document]
AdrienLE: This is actually pretty important so I added a clarifying comment in the next commit.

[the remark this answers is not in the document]
JoostHuizinga: Clear!

minerl/herobraine/env_spec.py
```diff
@@ -168,35 +168,28 @@ def is_from_folder(self, folder: str) -> bool:
...
+        return spaces.Dict({
             agent: spaces.Dict({
                 a.to_string(): a.space for a in self.actionables
             }) for agent in self.agent_names
-        }))
+        })
 
     def create_monitor_space(self):
-        return self._singlify(spaces.Dict({
+        return spaces.Dict({
```
MadcowD, line 188: Hmm, why are you removing the _singlify code? That should be breaking in the unit tests..

MadcowD (changes requested):

minerl/herobraine/env_spec.py
JoostHuizinga: The reason is that we want to be able to have an environment with a single agent while still using the multi-agent environment interface (which makes it much easier to write subsequent wrappers). However, _singlify only looks at self.agent_count, making it impossible to specify such an environment.

Perhaps the env_spec can have a separate parameter called `multi_agent_interface` (or something along those lines), which makes it possible to force a multi-agent interface in single-agent environments?

AdrienLE: What Joost is saying is exactly right: if we have environments that can return radically different formats depending on the number of agents, all the rest of our code will need to be special cased to handle that (I expect we'll often run sanity checks with a single agent), so it was simpler to just always return the multiagent format. I modified the unit tests that were broken by this accordingly.

If you prefer we can have a separate argument from agent_count handle this as Joost suggested.

brandonhoughton: What if we just set agent_count=None and that specified this singular env? Or is that overloading behaviors too much

AdrienLE: Will do the agent_count=None version, seems reasonable to me.

AdrienLE:
@MadcowD and @brandonhoughton, just made a change to still have _singlify when agent_count=None (and to make agent_count=None the default), does that work?

minerl/herobraine/env_spec.py
```diff
@@ -27,11 +27,12 @@ class EnvSpec(abc.ABC):
...
     U_FAKE_SINGLE_AGENT_ENTRYPOINT = 'minerl.env._fake:_FakeSingleAgentEnv'
 
-    def __init__(self, name, max_episode_steps=None, reward_threshold=None, agent_count=1):
+    def __init__(self, name, max_episode_steps=None, reward_threshold=None, agent_count=None):
         self.name = name
         self.max_episode_steps = max_episode_steps
         self.reward_threshold = reward_threshold
-        self.agent_count = agent_count
+        self.agent_count = 1 if agent_count is None else agent_count
+        self.is_single_agent = agent_count is None
```
JoostHuizinga, line 35: I think this would make more sense as a method:
```
def is_single_agent(self):
    return self.agent_count is None
```

That would make sure that the single-agent/multi-agent nature of the env_spec properly changes when the agent count is modified after construction, and it makes it much easier in the future if we want to ever move away from the `agent_count is None` structure.

JoostHuizinga (approved):
Looks good to me.

brandonhoughton (approved):

brandonhoughton:
I'm going to merge this now so I can move the code dir @MadcowD if you have additional concerns still I'm happy to refactor this later!

PR #414, opened 2020-10-23
remove xvfb from launchClient.sh

pzhokhov:
otherwise, not possible to run python with virtual display because of "DISPLAY" variable conflict


brandonhoughton (approved):

pzhokhov:
closed - my mistake, the nested xfvb-run calls do not conflict if used correctly

PR #417, opened 2020-10-26
multi-agent reward handlers 

pzhokhov:
a bunch of the java reward handlers compute the reward via subscribing to particular events via eventbus. When such events are generated server-side, reward for multi-agent environments can be misattributed (i.e. all the reward is accumulated by the client in the same java process as server, instead of going to the correct agent)

This PR:
- adds function to RewardBase that lets subclasses to check if the player that generated the reward is the same that instantiated the Reward class.  The logic is as follows - the Reward methods that are annottated with `SubscribeEvent` should either a) forward the reward to client via explicit message OR b) pass the check. 
- Fixes RewardForPossesingItem to send messages on obtaining and loosing the item 
- Deprecates RewardForCollectingItem

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -284,7 +284,7 @@ tests/logs/
 jenkins_home/
 
 #eclipse
-Malmo/Minecraft/*.launch
+minerl/Malmo/Minecraft/*.launch
 *minerl_watchers/
 caches/
 native/
--- a/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/RewardForPossessingItemImplementation.java
+++ b/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/RewardForPossessingItemImplementation.java
@@ -21,13 +21,22 @@ package com.microsoft.Malmo.MissionHandlers;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Map;
 
+import javax.xml.bind.DatatypeConverter;
+
+import com.microsoft.Malmo.MalmoMod;
+import com.microsoft.Malmo.MalmoMod.IMalmoMessageListener;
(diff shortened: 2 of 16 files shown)


.gitignore
```diff
@@ -284,8 +284,8 @@ tests/logs/
 jenkins_home/
 
 #eclipse
-Malmo/Minecraft/*.launch
+minerl/Malmo/Minecraft/*.launch
```
brandonhoughton, line 287: Nice catch =)

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/RewardForPossessingItemImplementation.java
```diff
@@ -267,5 +315,6 @@ public void getReward(MissionInit missionInit, MultidimensionalReward reward) {
     public void cleanup() {
         super.cleanup();
         MinecraftForge.EVENT_BUS.unregister(this);
+        MalmoMod.MalmoMessageHandler.deregisterForMessage(this, MalmoMessageType.SERVER_COLLECTITEM);
```
brandonhoughton, line 318: Should you deregister SERVER_DISCARDITEM here as well?

brandonhoughton (changes requested):

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/RewardForPossessingItemImplementation.java
pzhokhov: good point, thanks!

pzhokhov: done!

brandonhoughton (approved):

PR #418, opened 2020-10-27
Make it possible to specify health and food

AdrienLE:
Also adds an 'is_alive' observation which will be useful now that agents are easier to kill.

### DIFF
--- a/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Server/ServerStateMachine.java
+++ b/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Server/ServerStateMachine.java
@@ -21,24 +21,21 @@ package com.microsoft.Malmo.Server;
 
 import java.lang.reflect.Field;
 import java.math.BigDecimal;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Random;
+import java.util.*;
 
 import javax.xml.bind.JAXBElement;
 import javax.xml.bind.JAXBException;
 
 import net.minecraft.client.Minecraft;
-import net.minecraft.entity.EntityList;
+import net.minecraft.entity.*;
+import net.minecraft.entity.ai.attributes.AttributeModifier;
+import net.minecraft.entity.ai.attributes.IAttributeInstance;
 import net.minecraft.entity.player.EntityPlayerMP;
 import net.minecraft.item.ItemStack;
 import net.minecraft.launchwrapper.Launch;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.management.PlayerList;
+import net.minecraft.util.FoodStats;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.Vec3d;
 import net.minecraft.util.text.TextComponentString;
@@ -807,6 +804,16 @@ public class ServerStateMachine extends StateMachine
             return player;
         }
 
+        private void setMaxHealth(EntityLivingBase entity, float maxHealth) {
+            UUID MODIFIER_ID = UUID.fromString("d7efc338-70d7-42dd-afc0-b198a67c926e");
+            IAttributeInstance attribute = entity.getEntityAttribute(SharedMonsterAttributes.MAX_HEALTH);
+            attribute.removeModifier(MODIFIER_ID);
+            double toAdd = maxHealth - attribute.getAttributeValue();
+            // Note: the last argument being 0 means that this is an addition modifier.
+            // See https://minecraft.gamepedia.com/Attribute#Modifiers
+            attribute.applyModifier(new AttributeModifier(MODIFIER_ID, "Set Max Health", toAdd, 0));
+        }
+
         private void initialisePlayer(String username, String agentname)
         {
             System.out.println("Initializing player with username " + username + " and agentname " + agentname);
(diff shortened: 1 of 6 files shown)


minerl/Malmo/Schemas/Mission.xsd
```diff
@@ -173,6 +173,18 @@
...
+                        <xs:attribute name="health" type="xs:float" use="optional"/>
+                        <xs:attribute name="maxHealth" type="xs:float" default="20" />
+                    </xs:complexType>
+                </xs:element>
+                <xs:element name="StartingFood" minOccurs="0" maxOccurs="1">
+                    <xs:complexType>
+                        <xs:attribute name="food" type="xs:int" default="20" />
+                        <xs:attribute name="foodSaturation" type="xs:float" use="optional" />
+                    </xs:complexType>
+                </xs:element>
```
JoostHuizinga, line 187: Is it possible to to make all these values optional, just for flexibility?

minerl/herobraine/hero/handlers/agent/observations/lifestats.py
```diff
@@ -46,6 +47,18 @@ def xml_template(self) -> str:
...
+class _IsAliveObservation(LifeStatsObservation):
+    """
+    Handles is_alive observation. Initial value is True (alive)
+    """
+
+    def __init__(self):
+        keys = ['is_alive']
+        super().__init__(hero_keys=keys, univ_keys=keys,
+                         space=spaces.Box(low=False, high=True, shape=(), dtype=np.bool),
+                         default_if_missing=True)
```
JoostHuizinga, line 59: Is it true that if the observation is missing the value should be true, rather than false?

minerl/herobraine/hero/handlers/agent/start.py
```diff
@@ -93,3 +93,61 @@ def __init__(self, x, y, z, yaw):
...
+        For example:
+
+            starting_health = StartingHealthAgentStart(2.5)
+
+        Args:
+            max_health: The maximum amount of health the agent can have
+            health: The amount of health the agent starts with (max_health if not specified)
+        """
+        self.health = health
+        self.max_health = max_health
```
JoostHuizinga, line 124: I guess these defaults effectively make both values optional.

JoostHuizinga (commented):
Looks good, the only relevant comments are my questions about the MODIFIER_ID variable.

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Server/ServerStateMachine.java
```diff
@@ -21,24 +21,21 @@
...
 import java.lang.reflect.Field;
 import java.math.BigDecimal;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Random;
+import java.util.*;
```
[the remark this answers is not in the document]
AdrienLE: Actually I think PyCharm did it for me. Personally I don't feel strongly either way but I don't know whether we have style rules in that codebase so maybe I'll revert it.

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Server/ServerStateMachine.java
```diff
@@ -807,6 +804,16 @@ private EntityPlayerMP getPlayerFromUsername(String username)
             return player;
         }
 
+        private void setMaxHealth(EntityLivingBase entity, float maxHealth) {
+            UUID MODIFIER_ID = UUID.fromString("d7efc338-70d7-42dd-afc0-b198a67c926e");
```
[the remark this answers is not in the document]
AdrienLE: Will do, I think I used all caps because it's a constant and that's usually how I mark that in python, but I can changed that. The magic string is just a hardcoded random string since this thing needs a unique UUID per entity modifier for a given entity.

minerl/Malmo/Schemas/Mission.xsd
AdrienLE: Actually they are both optional: the first one because it has a default value (so you don't need to specify it) and the second one because it's marked as "optional". Basically the difference between "optional" and "default" is that "optional" seems to be the way to make it so that the default value is null.

minerl/herobraine/hero/handlers/agent/start.py
AdrienLE: To be clear this is also true in the xml.

minerl/herobraine/hero/handlers/agent/observations/lifestats.py
AdrienLE: Hmm, I think so. My assumption is that it's in case we haven't requested that this key be returned as an observation from Minecraft at all, in which case it's a better bet that the agent is alive rather than dead. Another way to see this is that default_if_missing for _LifeObservation is the max life rather than 0, which seems consistent with is_alive being true rather than false.

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Server/ServerStateMachine.java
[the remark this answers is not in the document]
JoostHuizinga: Wait, PyCharm, not IntelliJ?

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Server/ServerStateMachine.java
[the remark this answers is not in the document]
JoostHuizinga: Right, it is a local constant, got it. I guess I personally only use all-caps for global constants.

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Server/ServerStateMachine.java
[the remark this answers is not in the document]
AdrienLE: Yeah sorry, IntelliJ.

minerl/Malmo/Schemas/Mission.xsd
JoostHuizinga: Yeah, I realized that later in my review...

minerl/herobraine/hero/handlers/agent/start.py
JoostHuizinga: Yep, you can ignore that comment.

minerl/herobraine/hero/handlers/agent/observations/lifestats.py
JoostHuizinga: Fair. In that case, looks good to me!

JoostHuizinga (approved):
I would add a comment explaining the random string for the UUID, but looks good otherwise.

JoostHuizinga (approved):
No further comments

PR #420, opened 2020-10-28
control distance between spawned agents

pzhokhov:
Adds and implements NearAgent mission xml element that allows
 - agents to be placed within certain horizontal and vertical distance from each other
 - make agents see each other at the start

NOTE: it is not exactly clear how to generalize "LookingAt" for >2 agents case.

### DIFF
--- a/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Server/ServerStateMachine.java
+++ b/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Server/ServerStateMachine.java
@@ -888,6 +888,9 @@ public class ServerStateMachine extends StateMachine
                 double angle = rng.nextDouble() * 2.0 * Math.PI;
                 double Z = anchor.posZ + distance * Math.cos(angle);
                 double X = anchor.posX + distance * Math.sin(angle);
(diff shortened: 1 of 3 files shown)


minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Server/ServerStateMachine.java
```diff
@@ -809,6 +809,7 @@ private EntityPlayerMP getPlayerFromUsername(String username)
 
         private void initialisePlayer(String username, String agentname)
         {
+            System.out.println("Initializing player with username " + username + " and agentname " + agentname);
```
brandonhoughton, line 812: We should ideally log these things instead of relying on std out but since this is captured by the logger currently we can refactor all of the printing later

brandonhoughton (approved):

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Server/ServerStateMachine.java
pzhokhov: https://github.com/openai/openai/projects/2#card-48563180

PR #421, opened 2020-10-28
Fix smelting

pzhokhov:


minerl/Malmo/Minecraft/build.gradle
```diff
@@ -101,6 +101,7 @@ jar {
 
 // And add to the jvm args so that it is also loaded when running using gradle runClient:
 JavaExec exec = project.getTasks().getByName("runClient")
+// exec.jvmArgs(["-Dfml.coreMods.load=com.microsoft.Malmo.OverclockingPlugin","-Xmx2G", "-Xdebug", "-Xrunjdwp:transport=dt_socket,address=127.0.0.1:8888,server=y,suspend=n"])
```
pzhokhov, line 104: this option is to enable attaching of a debugger to a running minecraft instance

minerl/Malmo/Schemas/MissionHandlers.xsd
```diff
@@ -1636,13 +1636,13 @@
             <xs:documentation>
                 A command for simple smelting:
 
-                "{{{ smeltNearby carpet PINK}}}"
+                "{{{ nearbySmelt carpet PINK}}}"
```
JoostHuizinga, line 1639: Interesting, so this was the same bug as the one we had with craftNearby/nearbyCraft, but it seems we are choosing opposite solutions.

Note that I do not really have a strong opinion on this, and I think Brandon recommended that I go with "craftNearby", but I can also see an argument for "nearbyCraft" because that is the name of the object in Java.

Either way, I think our solution should be consistent.

minerl/Malmo/Minecraft/Minecraft_Server.launch
```diff
@@ -1 +0,0 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?><launchConfiguration type="org.eclipse.jdt.launching.localJavaApplication"><stringAttribute key="org.eclipse.jdt.launching.MAIN_TYPE" value="GradleStartServer"/><stringAttribute key="org.eclipse.jdt.launching.PROJECT_ATTR" value="Minecraft"/><stringAttribute key="org.eclipse.jdt.launching.WORKING_DIRECTORY" value="/Users/wguss/workspace/openai/minerl_data/minerl/minerl/env/Malmo/Minecraft/run"/><stringAttribute key="org.eclipse.jdt.launching.VM_ARGUMENTS" value=" -DFORGE_FORCE_FRAME_RECALC=true"/></launchConfiguration>
```
JoostHuizinga, line 1: What was the purpose of this file? Just an accidental configuration file for eclipse?

JoostHuizinga (changes requested):
This implements a solution that is the opposite from the solution we chose for crafting (we chose craftNearby over nearbyCraft). I think we should be consistent, and there is a good argument to be made for changing craftNearby to nearbyCraft.

minerl/Malmo/Minecraft/Minecraft_Server.launch
pzhokhov: Yes. Moreover, it was supposed to be included in the .gitignore, but due to a bug in the latter was not

minerl/Malmo/Schemas/MissionHandlers.xsd
pzhokhov: Good point, I did not compare it to craftNearby / nearbyCraft. Will fix to the craft standard.

brandonhoughton: I'm with changing craftNearby as well, the less ambiguity here the better

pzhokhov: uh I am confused now, which version would you guys prefer me to follow? Right now it is changed so that java commands are called smeltNearby and craftNearby, is that fine?

JoostHuizinga (approved):
I don't have a strong opinion on the craftNearby/nearbyCraft smeltNearby/nearbySmelt issue as long as it is consistent. The current pull request is consistent, so I think we can merge it.

PR #422, opened 2020-10-28
multi-threaded startup of java instances corresponding to different agents

pzhokhov:


minerl/env/_multiagent.py
```diff
@@ -545,9 +546,11 @@ def _setup_instances(self) -> None:
         """Sets up the instances for the environment 
         """
         num_instances_to_start = self.task.agent_count - len(self.instances)
-        self.instances.extend(
-            [self._get_new_instance() for _ in range(num_instances_to_start)]
-        )
+        instance_futures = []
+        with ThreadPoolExecutor(max_workers=num_instances_to_start) as tpe:
```
brandonhoughton, line 550: I have a vauge memory of this suffering from a race condition - I believe that it was resolved, so I'm going to say let's go ahead and merge this and I can investigate if it causes issues =)

brandonhoughton (approved):

PR #423, opened 2020-10-29
Add massive number of observations to MineRL

brandonhoughton:


minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Client/ClientStateMachine.java
```diff
@@ -25,17 +25,14 @@
 import java.math.BigDecimal;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
AdrienLE, line 28: Just curious: do we have any norms about when we import things individually vs using a * import?

brandonhoughton: No, convention of Malmo is to explicitly include every sub folder, where Intelij likes to automatically compact them after there are more than n (idk what n is) Happy to disable this if we prefer the other convention

minerl/Malmo/Minecraft/build.gradle
```diff
@@ -35,6 +36,10 @@ apply plugin: 'org.spongepowered.mixin'
 // For versions < 1.8
 // apply plugin: 'forge'
 
+apply plugin: 'java'
+sourceCompatibility = targetCompatibility = compileJava.sourceCompatibility = compileJava.targetCompatibility = '1.8' // Need this here so eclipse task generates correctly.
```
brandonhoughton, line 40: Build against Java 1.8

minerl/data/pipeline/make_minecrafts.py
```diff
@@ -12,7 +12,103 @@
 from minerl.data.util.constants import NUM_MINECRAFTS
 
 MINECRAFT_TEMPLATE = os.path.abspath(J(os.path.dirname(__file__), '..', 'assets', 'template_minecraft'))
-DOTMINECRAFT = os.path.abspath(J(BASE_DIR, '.minecraft'))
+DOTMINECRAFT = os.path.abspath(J(BASE_DIR, 'testcraft'))
```
brandonhoughton, line 15: Should be removed

minerl/data/pipeline/make_minecrafts.py
```diff
@@ -165,7 +261,11 @@ def download(url, file_name):
         shutil.copytree(MINECRAFT_TEMPLATE, target_mc_name)
         xauth = os.path.join(target_mc_name, 'xauth')
         with open(os.path.join(target_mc_name, 'launch.sh'), 'w') as f:
-            f.write(LAUNCH_STR.format(DOTMINECRAFT, cracked_libs, target_mc_name, i + 20, xauth))
+            if False:
```
brandonhoughton, line 264: Should be removed

minerl/data/pipeline/merge.py
```diff
@@ -184,9 +184,12 @@ def merge_stream(stream_name):
     # 2. Merge files.
     # 3. 7z files
     # 4. Place ifles
+    tempdir = None
     try:
-        target_name = J(MERGED_DIR, "{}.mcpr".format(stream_name))
+        # tempdir = tempfile.mkdtemp()
```
brandonhoughton, line 189: Should be removed

minerl/data/util/constants.py
```diff
@@ -30,7 +30,7 @@ def module_path():
 
 RENDERERS_DIR = os.path.expanduser(
     J('~', 'renderers'))
-NUM_MINECRAFTS = 28
+NUM_MINECRAFTS = 2
```
brandonhoughton, line 33: Remove rendering changes from this pr

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Client/ClientStateMachine.java
AdrienLE: Personally I think it's fine to use `*`, but I know Joost said he didn't super like it. Probably just a super low pri thing.

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Utils/JSONWorldDataHelper.java
```diff
@@ -112,33 +116,63 @@ public GridDimensions(int xMargin, int zMargin) {
...
-        */
+            for(StatBase stat : StatList.ALL_STATS) {
+                // For MineRL, split over . and convert all camelCase to snake_case
+                String[] stat_fields = stat.statId.split("\\.");
+                JsonObject head = json;
+                for (String unformatted_token : stat_fields) {
+                    String token = CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_UNDERSCORE, unformatted_token);
+                    // Last element is a leaf
+                    if (unformatted_token.equals(stat_fields[stat_fields.length - 1])) {
+                        // BAH map drop stat to items_dropped to prevent hash collision in dict keys
```
AdrienLE, line 149: What does `BAH` mean?

minerl/env/_multiagent.py
```diff
@@ -530,7 +531,8 @@ def _setup_agent_xmls(self, ep_uid: str) -> List[etree.Element]:
                         <MaxPlayers>{}</MaxPlayers>
                     </HumanInteraction>""".format(self.interact_port, self.max_players))
                 # Update the xml
-                ss = agent_xml_etree.find(".//" + self.ns + 'ServerSection')
+                namespace = '{http://ProjectMalmo.microsoft.com}'
```
AdrienLE, line 534: Why is this no longer `self.ns` or just the `NS` global constant which seems to contain the same value?

minerl/env/_multiagent.py
```diff
@@ -545,9 +547,12 @@ def _setup_instances(self) -> None:
...
         num_instances_to_start = self.task.agent_count - len(self.instances)
-        self.instances.extend(
-            [self._get_new_instance() for _ in range(num_instances_to_start)]
-        )
+        instance_futures = []
+        if num_instances_to_start > 0:
+            with ThreadPoolExecutor(max_workers=num_instances_to_start) as tpe:
+                for _ in range(num_instances_to_start):
+                    instance_futures.append(tpe.submit(self._get_new_instance))
+        self.instances.extend([f.result() for f in instance_futures])
```
AdrienLE, line 555: Is this to launch instances faster?

minerl/herobraine/hero/handlers/agent/observations/location_stats.py
```diff
@@ -0,0 +1,158 @@
...
+class _BiomeIDObservation(_FullStatsObservation):
+    def to_hero(self, x):
+        for key in self.hero_keys:
+            x = x[key]
+        return np.eye(40)[x]
+
+    def __init__(self):
+        super().__init__(key_list=['biome_id'],
+                         space=spaces.Discrete(40),
+                         default_if_missing=np.eye(40)[0])
```
AdrienLE, line 111: Is it because there are 40 biomes? Is there a good way to handle a change in the number of biomes?

AdrienLE: Also, probably not a big deal but unless I'm mistaken this creates an entire 40x40 matrix to only extract the x'th row. Is there a standard way to do this (a onehot function we already have on hand perhaps) that doesn't allocate and initialize 40 times more data than it needs to?

minerl/herobraine/hero/handlers/agent/observations/location_stats.py
```diff
@@ -0,0 +1,158 @@
...
+    def __init__(self):
+        super().__init__(key_list=['light_level'],
+                         space=spaces.Box(low=0.0, high=15, shape=(), dtype=np.int),
+                         default_if_missing=15)
+
+class _IsRainingObservation(_FullStatsObservation):
+    def to_hero(self, x):
+        for key in self.hero_keys:
+            x = x[key]
+        return np.eye(2)[x]
```
AdrienLE, line 141: For these boolean observations, couldn't we just use a single element that could be 0 or 1 instead of one-hot encoding it?

minerl/herobraine/hero/handlers/agent/observations/location_stats.py
```diff
@@ -0,0 +1,158 @@
...
+    def __init__(self):
+        super().__init__(key_list=['is_raining'],
+                         space=spaces.Discrete(2),
+                         default_if_missing=np.eye(2)[0])
+
+class _CanSeeSkyObservation(_FullStatsObservation):
+    def to_hero(self, x):
+        for key in self.hero_keys:
+            x = x[key]
+        return np.eye(2)[x]
```
AdrienLE, line 152: Same comment as above about boolean observations.

AdrienLE (changes requested):
Left some minor comments. Don't necessarily all require changes but at least responses. 

We've been talking about having tests on the OpenAI side. Should we also have more tests on the minerl side? Probably not crucial yet but perhaps worth discussing soon.

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Utils/JSONWorldDataHelper.java
brandonhoughton: My initials =) Maybe I should change given we have git annotations

minerl/env/_multiagent.py
brandonhoughton: I don't remember changing this - I think it was pulled from another branch - reverting

minerl/herobraine/hero/handlers/agent/observations/location_stats.py
brandonhoughton: Yep!

AdrienLE: Does that mean you're going to change it?

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Utils/JSONWorldDataHelper.java
AdrienLE: I see. Not necessarily crucial to change, maybe I just need to get used to it.

minerl/herobraine/hero/handlers/agent/observations/location_stats.py
brandonhoughton: You're correct, I stole this from another project. It does use more space but it is fewer lines than making a vector and assigning the nth element to 1. Since there could be more biomes later I think I will just refactor all of these to be int. In general we preferred showing the independence of consecutive values by encoding them as a one-hot from MineRL because not everyone wraps things.

brandonhoughton: Nevermind - the discrete space dictates the type here. There are currently biomes with IDs as large as ~400 ... they don't appear in the base game but could be enabled later

minerl/env/_multiagent.py
brandonhoughton: Yes this launches them in parallel

AdrienLE (approved):
Works for me!

PR #424, opened 2020-10-30
set line length to 100 instead of 79 in pep8speaks

AdrienLE:


JoostHuizinga (commented):
Fine by me, but I also feel like I am not really the person to judge this change.

brandonhoughton (approved):

brandonhoughton:
I personally prefer longer lines, though the code was previously linted at 79 so it could be a touch confusing. We can just use 100 moving forward (or switch to 88 to mirror black)

PR #428, opened 2020-11-05
prevent agent skin blinking

pzhokhov:
The function "isSlimSkin" is actually being called more than once (I assume every time client decides it needs to render the other player?). Because we are currently using a random boolean as a result, the agent "blinks" between two skins (alex and steve) - can be reproduced in interactive mode with 2 agents. 
This fix chooses the random number once and adds it to the UUID when deciding which skin to use. It fixes blinking, however, when there are more than 2 agents, the results are not consistent between the observers.

### DIFF
--- a/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Mixins/MixinRandomSkinTexture.java
+++ b/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Mixins/MixinRandomSkinTexture.java
@@ -17,14 +17,18 @@ import net.minecraft.client.resources.DefaultPlayerSkin;
 @Mixin(DefaultPlayerSkin.class)
 public abstract class MixinRandomSkinTexture {
     // Randomize the skin for agents ignoring UUID
-    private static Boolean isSlim = null;
+    private static Integer skinSeed = null;
     @Inject(method = "isSlimSkin", at = @At("HEAD"), cancellable = true)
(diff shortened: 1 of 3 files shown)


brandonhoughton:
Thanks for fixing this - seems like we should prioritize teams now and have team assignment be random so everyone agrees on the skin?

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Mixins/MixinRandomSkinTexture.java
```diff
@@ -3,24 +3,32 @@
...
     // Randomize the skin for agents ignoring UUID
+    private static Integer skinSeed = null;
     @Inject(method = "isSlimSkin", at = @At("HEAD"), cancellable = true)
     private static void isSlimSkin(UUID playerUUID, CallbackInfoReturnable<Boolean> cir){
-        Random rand = new Random();
-        cir.setReturnValue(rand.nextBoolean());
+        if (skinSeed == null) {
+            // TODO now this is not blinking, and randomized, but not consistent between
+            // clients. Ideally, we should somehow sync this between clients -
+            // is there any pseudo-random number in the minecraft world we
```
brandonhoughton, line 26: We have UUID of the agent, we could just salt the UUID with some server wide information like map seed?

pzhokhov: yeah, using map seed sounds great!

brandonhoughton (approved):

PR #438, opened 2021-01-05
Fix multiagent crashes

AdrienLE:
Fix crashes in Minecraft due to three issues:
    - Loss of connection from the client to the server weren't handled properly. Now makes it so that the client uses its old data when the connection is lost instead of crashing.
    - A race condition between the client going idle and the server asking the clients to stop. Now the server repeatedly asks the client to stop in case the message was lost.
    - When one of the client dies, it tries to trigger end of episode. This is bad for multiagent because we might want to keep the episode going and also it seemed to cause issues with the clients getting desynchronized (one thinks the episode has ended, the other does not), causing crashes. Now we don't end episodes on the Java side if a client dies and we are in multiagent.

### DIFF
--- a/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Client/MalmoEnvServer.java
+++ b/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Client/MalmoEnvServer.java
@@ -476,8 +476,8 @@ public class MalmoEnvServer implements IWantToQuit {
             profiler.endSection(); //cmd
             profiler.startSection("clientTick");
 
-            // TimeHelper.SyncManager.debugLog("[MALMO_ENV_SERVER] <STEP> Received: " + actions);
-            // TimeHelper.SyncManager.debugLog("[MALMO_ENV_SERVER] <STEP> Requesting tick. Should use previous env state: " + Boolean.toString(shouldUsePreviousEnvState));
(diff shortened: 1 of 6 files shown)


minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Client/MalmoEnvServer.java
```diff
@@ -474,13 +477,18 @@ private synchronized void stepClientSync(String command, Socket socket, DataInpu
             profiler.startSection("clientTick");
 
             // TimeHelper.SyncManager.debugLog("[MALMO_ENV_SERVER] <STEP> Received: " + actions);
```
brandonhoughton, line 479: I think in general we have cleaner code when removing old debug statements. In this case they may be helpful for diagnosing new bugs if they appear - I could go either way

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/StateMachine.java
```diff
@@ -20,6 +20,9 @@
 package com.microsoft.Malmo;
 
 import java.util.ArrayList;
+import java.util.concurrent.locks.Lock;
```
brandonhoughton, line 23: I added this but I believe this import is unused

brandonhoughton (approved):

brandonhoughton:
Added some minor comments - looks good!

PR #445, opened 2021-01-28
Update gradle for setup_intellij.sh script

brandonhoughton:
Current version of gradle is not supported by intellij and also breaks the setup script.


minerl/Malmo/Minecraft/gradlew
```diff
@@ -6,12 +6,30 @@
 ##
 ##############################################################################
 
-# Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS=""
```
JoostHuizinga, line 10: Why has this statement moved down? I assumed it was near the top of the file such that you would quickly see it you wanted to add custom options.

minerl/Malmo/Minecraft/gradlew.bat
```diff
@@ -8,14 +8,14 @@
 @rem Set local scope for the variables with windows NT shell
 if "%OS%"=="Windows_NT" setlocal
 
-@rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-set DEFAULT_JVM_OPTS=
```
JoostHuizinga, line 12: Same comment here. As in, it looks like the code below should not be affected by DEFAULT_JVM_OPTS at all. Was it affected?

JoostHuizinga (approved):
Looks good to me (the only comments I have are super minor). That said, I don't really know that much about the gradle setup, so I would not be able to spot any subtle errors.

PR #449, opened 2021-02-02
human action space + enabled GUI

pzhokhov:
- adds a "LowLevelInput" mission flag that allows malmo input to be injected at mouse and keyboard level (instead of events level)
- when lowLevelInput is on, in-game GUIs are enabled
- new env specs - HumanControls (base spec class), HumanSurvival, and EquipWeapon. EquipWeapon is task to equip iron axe that is placed somewhere in the inventory using human inputs. The reward and done signal for equip weapon have to be implemented via a wrapper on the python side.


minerl/Malmo/Minecraft/build.gradle
```diff
@@ -105,8 +105,9 @@ jar {
 
 // And add to the jvm args so that it is also loaded when running using gradle runClient:
 JavaExec exec = project.getTasks().getByName("runClient")
-// exec.jvmArgs(["-Dfml.coreMods.load=com.microsoft.Malmo.OverclockingPlugin","-Xmx2G", "-Xdebug", "-Xrunjdwp:transport=dt_socket,address=127.0.0.1:8888,server=y,suspend=n"])
-exec.jvmArgs(["-Dfml.coreMods.load=com.microsoft.Malmo.OverclockingPlugin","-Xmx2G"])
+exec.jvmArgs(["-Dfml.coreMods.load=com.microsoft.Malmo.OverclockingPlugin","-Xmx2G", "-Xdebug", "-Dmixin.debug=true"])
```
ilge, line 108: just want to make sure the debug flags won't impact experiment envs?

pzhokhov: pretty sure it will not (it's the debug port flag that fails when there is more than one java process on a machine)

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Client/ClientStateMachine.java
```diff
@@ -1874,9 +1874,15 @@ protected void onMissionStarted()
 
             // Force brightness setting
             Minecraft.getMinecraft().gameSettings.gammaSetting = (float) 2.0;
```
brandonhoughton, line 1876: A gamma setting of 2 is brighter than playable without editing the options.txt manually (e.g. the highest in game via the brightness slider)

brandonhoughton: May also need to look into fov as it is beyond quake-pro in default minerl

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Client/MalmoModClient.java
```diff
@@ -243,20 +245,8 @@ public void onRightClickEvent(PlayerInteractEvent.RightClickBlock event){
         }
     }
 
-    /**
-     * Event listener that logs when agents open gui windows
-     * @param event the captured event
-     */
-    @SideOnly(Side.CLIENT)
```
brandonhoughton, line 250: This was originally here to debug if we were opening a gui window unintentionally, probably OK to remove

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Client/MalmoModClient.java
```diff
@@ -208,43 +208,6 @@ public void onPressed()
         this.keyManager = new KeyManager(settings, extraKeys);
     }
 
-    /**
```
brandonhoughton, line 209: How will we prevent locking out non-gui agents from moving?

pzhokhov: oh you make a good point, let me restore this bit of code. Deleted it accidentally with the debug function ^_^

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Client/ClientStateMachine.java
pzhokhov: good point, but I'd rather get a wider team buy-in before changing these values, since that would affect everyone's experiments. Currently in this PR the behavior (without lowLevelInput flag) should not change.

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/CameraCommandsImplementation.java
```diff
@@ -46,28 +48,34 @@ public void setOverriding(boolean b) {
...
-                this.currentYaw = player.rotationYaw;
-                this.currentPitch = player.rotationPitch;
+                        this.currentYaw = player.rotationYaw;
+                        this.currentPitch = player.rotationPitch;
+                    }
+                }
+                return true;
             }
         } catch (NumberFormatException e) {
             System.out.println("ERROR: Malformed parameter string (" + parameter + ") - " + e.getMessage());
```
brandonhoughton, line 76: This is ignoring number format exceptions then? That is expected right?

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Mixins/MixinMouse.java
```diff
@@ -0,0 +1,113 @@
...
+    public static void poll() {
+
+    }
+
+    @Overwrite(remap = false)
+    public static void setGrabbed(boolean grabbed) {
+        FakeMouse.setGrabbed(grabbed);
+    }
+//
+//    @Overwrite(remap = false)
```
brandonhoughton, line 26: No-longer needed?

minerl/Malmo/Schemas/Mission.xsd
```diff
@@ -235,6 +235,14 @@
                         <xs:attribute name="foodSaturation" type="xs:float" use="optional" />
                     </xs:complexType>
                 </xs:element>
+                <xs:element name="LowLevelInputs" type="xs:boolean" default="false" minOccurs="0" maxOccurs="1">
+                    <xs:annotation>
+                        <xs:documentation>
+                            If set to true, inputs to malmo are injected instead of keyboard and mouse (instead of higher level commands)
```
brandonhoughton, line 241: Maybe "If set to true, inputs to malmo are disabled and instead given via a virtual keyboard and mouse (instead of higher level commands)"

brandonhoughton (approved):
Left a few more comments but otherwise looks good to me!

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/CameraCommandsImplementation.java
pzhokhov: my understanding is that, just like prior to the change, in case of NumberFormatExceptions, we print an error message and return false (which, AFAIK, means that the command will be passed further down the chain)

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/CameraCommandsImplementation.java
```diff
@@ -82,19 +90,18 @@ protected boolean onExecute(String verb, String parameter, MissionInit missionIn
...
                     if(this.currentYaw == -10000 & this.currentPitch == -10000){
-                        
-
                         this.currentYaw = player.rotationYaw;
                         this.currentPitch = player.rotationPitch;
                     }
-                    player.setPositionAndRotation(player.posX, player.posY, player.posZ, this.currentYaw, this.currentPitch);
+                    // TODO peterz: I am pretty sure this code is deprecated and does not match the docstring
+                    // i.e. no smooth panning is happening here
+                    // player.setPositionAndRotation(player.posX, player.posY, player.posZ, this.currentYaw, this.currentPitch);
```
pzhokhov, line 104: seems to be causing extra flakiness in starting position?

PR #454, opened 2021-02-12
make it possible to tell which agent is which in create agent start

AdrienLE:
The method is a bit hacky (creating a stateful attribute to contain which agent is being processed), but I thought doing it this way would make it possible to not break existing code, vs adding a new argument to that function that is inherited all over the place.

### DIFF
--- a/minerl/herobraine/env_spec.py
+++ b/minerl/herobraine/env_spec.py
@@ -53,7 +53,9 @@ class EnvSpec(abc.ABC):
 
         # after create_server_world_generators(), because it will see python generated map
         # to pick a good location
-        self.agent_start = [self.create_agent_start() for _ in range(self.agent_count)]
+        self.agent_start = []
+        for self.current_agent in range(self.agent_count):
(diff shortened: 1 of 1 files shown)


minerl/herobraine/env_spec.py
```diff
@@ -53,7 +53,9 @@ def reset(self):
 
         # after create_server_world_generators(), because it will see python generated map
         # to pick a good location
-        self.agent_start = [self.create_agent_start() for _ in range(self.agent_count)]
```
brandonhoughton, line 56: I think this is equivalent to self.agent_start = [self.create_agent_start() for self.current_agent in range(self.agent_count)]

brandonhoughton (approved):

minerl/herobraine/env_spec.py
AdrienLE: Yeah good point that would probably work too. Somehow assigning a `self.<something>` made me feel like it wouldn't work so I switched it lol. Maybe not worth the effort to change though.

PR #458, opened 2021-02-23
make gui scale configurable from python

pzhokhov:


minerl/herobraine/env_specs/human_controls.py
```diff
@@ -54,4 +54,4 @@ def create_monitors(self) -> List[TranslationHandler]:
         return []  # No monitors by default!o
 
     def create_agent_start(self) -> List[Handler]:
-        return [handlers.LowLevelInputsAgentStart()]
+        return [handlers.LowLevelInputsAgentStart(), handlers.GuiScale(5)]
```
ilge, line 57: what are the different defaults 5.0 vs 4.0 in here and below?

PR #460, opened 2021-03-02
Added damage source observation to MineRL

brandonhoughton:


brandonhoughton:
TODO observe all damage incoming not just on death

brandonhoughton:
TODO print the damage so we can debug agents

.pep8speaks.yml
```diff
@@ -1,2 +1,2 @@
 pycodestyle:
-    max-line-length: 100  # Default is 79 in PEP 8
```
shwang: I'd be inclined to keep at 100: 120 is harder for multiple split screens on my computer.

.gitignore
```diff
@@ -292,3 +292,7 @@ wrapper/
 
 # Pytest artifacts
 tests/*.lock
+
+
+# Minecraft
+minerl/Malmo/Minecraft/run
```
shwang, line 298: Out of curiosity, how is this directory or file generated?

.pep8speaks.yml
brandonhoughton: Sure - I'll push a refractor to satisfy pep8 soon

.gitignore
brandonhoughton: Hmm true this is not well documented - it came from the dedicated server

minerl/Malmo/Schemas/MissionHandlers.xsd
```diff
@@ -2283,6 +2283,17 @@
         <xs:complexType/>
     </xs:element>
 
+    <xs:element name="ObservationFromDamage">
+        <xs:annotation>
+            <xs:documentation>
+                When present, the mod will return a JSON object called "damage_source", containing the observation
+                "is_dead" telling if the player is dead. Additionally, if there was an unprocessed damage event, there
+                are a number .
```
shwang: Looks like this comment is incomplete

.gitignore
shwang: Gotcha. Is that the buildkite server?

minerl/herobraine/hero/mission.xml.j2
```diff
@@ -60,6 +60,11 @@
             {{ x }}
             {% endfor %}
 
+            <!--  Monitors     -->
+
+            {% for x in get_consolidated_xml((monitors)) %}
+            {{ x }}
+            {% endfor %}
```
shwang, line 67: Oh interesting -- does this mean that Monitors were not previously activated on `dev`, if they were parts of EnvSpecs?

brandonhoughton: Yep! But they were processed if they shared obs so)

.gitignore
brandonhoughton: It's a new feature we're working on to use a full minecraft server rather than the integrated one that comes with clients

minerl/herobraine/hero/handlers/translation.py
```diff
@@ -62,6 +63,8 @@ def __init__(self,
         self.hero_keys = hero_keys
         self.univ_keys = univ_keys
         self.default_if_missing = default_if_missing
+        self.ignore_missing = ignore_missing
```
shwang, line 71: Could you add a docstring for `ignore_missing` here?

minerl/herobraine/env_specs/treechop_specs.py
```diff
@@ -26,12 +26,28 @@
...
+"useMineShafts":false,"useTemples":false,"useMonuments":false,"useMansions":false,"useRavines":false,\
+"useWaterLakes":false,"waterLakeChance":4,"useLavaLakes":false,"lavaLakeChance":80,"useLavaOceans":false,\
+"fixedBiome":4,"biomeSize":4,"riverSize":1,"dirtSize":33,"dirtCount":10,"dirtMinHeight":0,"dirtMaxHeight":256,\
+"gravelSize":33,"gravelCount":8,"gravelMinHeight":0,"gravelMaxHeight":256,"graniteSize":33,"graniteCount":10,\
+"graniteMinHeight":0,"graniteMaxHeight":80,"dioriteSize":33,"dioriteCount":10,"dioriteMinHeight":0,\
+"dioriteMaxHeight":80,"andesiteSize":33,"andesiteCount":10,"andesiteMinHeight":0,"andesiteMaxHeight":80,\
+"coalSize":17,"coalCount":20,"coalMinHeight":0,"coalMaxHeight":128,"ironSize":9,"ironCount":20,"ironMinHeight":0,\
+"ironMaxHeight":64,"goldSize":9,"goldCount":2,"goldMinHeight":0,"goldMaxHeight":32,"redstoneSize":8,"redstoneCount":8,\
+"redstoneMinHeight":0,"redstoneMaxHeight":16,"diamondSize":8,"diamondCount":1,"diamondMinHeight":0,\
+"diamondMaxHeight":16,"lapisSize":7,"lapisCount":1,"lapisCenterHeight":16,"lapisSpread":16}"""
```
AdrienLE: This change doesn't seem to make sense, it looks like backslashes were added everywhere lines break somehow, but I don't think that's the appropriate way to do it either in strings or in comments, no?

AdrienLE: Let me know if I'm not understanding something of course.

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/FileBasedPerformanceProducerImplementation.java
```diff
@@ -54,13 +58,42 @@ public boolean parseParameters(Object params)
...
             if(currentEpisodeJson != null && statusJson != null){
                 currentEpisodeJson.getAsJsonArray("rewards").add(new JsonPrimitive(reward));
                 currentEpisodeJson.addProperty("numTicks", 
                     currentEpisodeJson.get("numTicks").getAsLong() + 1L);
-                
+
+                // Update the damage type
+                if (info != null && info.has("damage_source")) {
+                    // Record damage type
+                    if (info.get("damage_source").getAsJsonObject().has("damage_type")){
```
shwang: suggestion: Refactor to have local variable `JsonObject dmgSource = info.get("damage_source").getAsJsonObject()` to reuse between if conditionals

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
```diff
@@ -0,0 +1,131 @@
...
+    private DamageSource damageSource;
+    private float damageAmount;
+    private EntityLivingBase entity;
+    private boolean hasDied = false;
+
+
+	@Override
+	public void prepare(MissionInit missionInit) {
+        MinecraftForge.EVENT_BUS.register(this);
+    }
```
shwang, line 37:
```suggestion
      }
```

spacing nit :)

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
```diff
@@ -0,0 +1,131 @@
...
+            this.hasDied = true;
+        }
+    }
+    
+    @SubscribeEvent
+    public void onDamage(LivingHurtEvent event)
+    {
+        if (event.getEntityLiving().equals(Minecraft.getMinecraft().player)) {
+            if (this.damageSource != null) {
+                System.out.println("Warning skipped damage event - multiple damage events in one tick!");
```
shwang, line 64: I see -- so the reason that landing in this if branch means that there are multiple damage events in one tick is because we expect  `writeObservationsToJSON` to be called every tick, which resets `this.damageSource` to `null`?

If that's true, then could you add that as a comment above the warning log?

shwang: Mainly to make the implication immediately obvious here

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    private void resetDamage(){
+	    this.damageAmount = 0;
+	    this.damageSource = null;
+	    this.entity = null;
+	    this.hasDied = false;
+    }
+
+	@Override
```
shwang:
```suggestion
    @Override
```

minerl/herobraine/env_specs/treechop_specs.py
brandonhoughton: I was trying to make the linter happy - in python if there is no space after the backslash it should join the string without additional spaces. That being said it can be broken by adding a silly little space somewhere so perhaps we leave linters angry

minerl/Malmo/Schemas/MissionHandlers.xsd
brandonhoughton: Got it

minerl/herobraine/hero/handlers/translation.py
brandonhoughton: got that

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
brandonhoughton: Right - this was left as a debugging artifact - I'll clarify in the comments. The issue is I expect that in rare cases you could receive multiple damage events per tick - perhaps I should see if this is possible first

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public void writeObservationsToJSON(JsonObject json, MissionInit missionInit)
+    {
+        EntityPlayerSP player = Minecraft.getMinecraft().player;
+        JsonObject damage_json = new JsonObject();
+        json.addProperty("is_dead", player.isDead);
+        json.addProperty("living_death_event_fired", this.hasDied);
+        this.hasDied = false;
```
shwang:
```suggestion
```

Should already be covered by `resetDamage()`

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
```diff
@@ -0,0 +1,131 @@
...
+            damage_json.addProperty("is_damage_absolute", this.damageSource.isDamageAbsolute());
+            damage_json.addProperty("is_fire_damage", this.damageSource.isFireDamage());
+            damage_json.addProperty("is_magic_damage", this.damageSource.isMagicDamage());
+            damage_json.addProperty("is_difficulty_scaled", this.damageSource.isDifficultyScaled());
+            damage_json.addProperty("is_explosion", this.damageSource.isExplosion());
+            damage_json.addProperty("is_projectile", this.damageSource.isProjectile());
+            damage_json.addProperty("is_unblockable", this.damageSource.isUnblockable());
+            damage_json.addProperty("death_message", this.damageSource.getDeathMessage(player).getUnformattedText());
+
+            if (this.damageSource.getEntity() != null) {
```
shwang, line 111: suggestion: Could shorten by saving `this.damageSource.getEntity()` as a local variable

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/FileBasedPerformanceProducerImplementation.java
```diff
@@ -36,6 +36,10 @@ public void prepare(MissionInit missionInit) {
         currentEpisodeJson = new JsonObject();
         currentEpisodeJson.addProperty("numTicks", 0);
         currentEpisodeJson.addProperty("environment", envName);
+        currentEpisodeJson.addProperty("damageAmount", 0);
```
shwang, line 39: Ah, so values saved to `currentEpisodeJson` are readable by AICrowd?

minerl/herobraine/hero/handlers/agent/observations/damage_source.py
```diff
@@ -0,0 +1,82 @@
...
+__all__ = ['ObservationFromDamageSource']
+
+
+class ObservationFromDamageSource(TranslationHandlerGroup):
+    """
+    Includes the most recent damage event including the amount, type, location, and many other properties
+    """
+
+    def xml_template(self) -> str:
+        return str("""<ObservationFromDamage/>""")
```
shwang:
```suggestion
        return "<ObservationFromDamage/>"
```

minerl/herobraine/hero/handlers/agent/observations/damage_source.py
```diff
@@ -0,0 +1,82 @@
...
+    def to_string(self) -> str:
+        return "damage_source"
+
+    def __init__(self):
+        super(ObservationFromDamageSource, self).__init__(
+            handlers=[
+                _HungerDamage(),
+                _DamageAmount(),
+                _IsDead(),
+                # _DamageType(),
```
shwang, line 33: Could you add a comment explaining why these handlers are removed for now?

shwang: Oh, I see, they aren't implemented yet. Maybe add another TODO on line 30? (I see you already have one at the bottom of the file)

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
```diff
@@ -0,0 +1,131 @@
...
+                JsonArray entity_equipment = new JsonArray();
+                // TODO do we need to mark the equipment slot of this armor?
+                for (ItemStack item : this.damageSource.getEntity().getEquipmentAndArmor()) {
+                    if (item.getItem().getRegistryName() != null) {
+                        JsonElement jitem = new JsonPrimitive(item.getItem().getRegistryName().toString());
+                        entity_equipment.add(jitem);
+                    }
+                }
+                damage_json.add("damage_entity_equipment", entity_equipment);
+            } if (this.damageSource.getDamageLocation() != null) {
```
shwang:
```suggestion
            }
            if (this.damageSource.getDamageLocation() != null) {
```

(since this isn't an else if)

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
```diff
@@ -0,0 +1,131 @@
...
+	@Override
+	public void prepare(MissionInit missionInit) {
+        MinecraftForge.EVENT_BUS.register(this);
+      }
+
+	@Override
+	public void cleanup() {}
+
+    @SubscribeEvent
+    public void onClientTick(LivingDeathEvent event)
```
shwang: Should this be called onDeath instead of onTick?

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
```diff
@@ -0,0 +1,131 @@
...
+ */
+public class ObservationFromDamageImplementation extends HandlerBase implements IObservationProducer
+{
+    private DamageSource damageSource;
+    private float damageAmount;
+    private EntityLivingBase entity;
+    private boolean hasDied = false;
+
+
+	@Override
```
shwang: Looks like this section has different spacing than the sections that follow. Starts with tabs instead of spaces?

shwang (changes requested):
It will be super helpful to see damage and death sources, or to be able to log and track these soon!

I got pretty detailed with suggestions here -- let me know if this is a helpful level or what a helpful level is, and I can adjust accordingly :)


I have a question involving the full damage dict -- it looks like we are only saving a subset of the damage dict to `PerformanceHelper.getOutDir()`. I didn't dive too deeply into what's going on when it comes to the Gym environment info dicts. Is it the case that the full damage source dict is available as `info` from `gym.step()` for all MineRL environment now?

shwang:
Merging in `dev` so that CircleCI can run properly (gets the necessary CircleCI config from #552)

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/FileBasedPerformanceProducerImplementation.java
brandonhoughton: Yep! If you want to include more things we can do anything in the dict I just wanted to be a bit more focused. Maybe we should point these out so WandB folks can integrate them

minerl/herobraine/hero/handlers/agent/observations/damage_source.py
brandonhoughton: yup!

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
brandonhoughton: Huhh good catch

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
brandonhoughton: should be fixed by reformatting

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
brandonhoughton: I think I got this with the reformatting

brandonhoughton:
CI is failing because we removed headless as the default - trying to add it back but I don't seem to have set the env var properly

brandonhoughton:
CI is happy now that we re-enabled headless mode on Buildkite @shwang want to take another look? Note some code changes are not showing in the comment section for me I had to try and edit the file before I saw the actual code (not sure why the dsync here)

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/FileBasedPerformanceProducerImplementation.java
shwang: nice

shwang:
I'll take another look at this tomorrow morning

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+
+    @SubscribeEvent
+    public void onDamage(LivingHurtEvent event) {
+        if (event.getEntityLiving().equals(Minecraft.getMinecraft().player)) {
+            if (this.damageSource != null) {
+                this.damageSourceReplaced = true;
+                // If multiple LivingHurtEvents pertain to the player, we keep the most recent one and record the fact
+                // that we are skipping a damage event in leiu of building an array based reporting system
```
shwang:
```suggestion
                // that we are skipping a damage event in lieu of building an array based reporting system
```

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
```diff
@@ -0,0 +1,142 @@
...
+            damage_json.addProperty("is_fire_damage", this.damageSource.isFireDamage());
+            damage_json.addProperty("is_magic_damage", this.damageSource.isMagicDamage());
+            damage_json.addProperty("is_difficulty_scaled", this.damageSource.isDifficultyScaled());
+            damage_json.addProperty("is_explosion", this.damageSource.isExplosion());
+            damage_json.addProperty("is_projectile", this.damageSource.isProjectile());
+            damage_json.addProperty("is_unblockable", this.damageSource.isUnblockable());
+            damage_json.addProperty("death_message", this.damageSource.getDeathMessage(player).getUnformattedText());
+
+            if (this.damageSource.getEntity() != null) {
+                Entity entity = this.damageSource.getEntity();
```
shwang, line 112: FYI -- it looks like `this.entity` is unused and overshadowed by this local variable here.

I would recommend renaming one of these guys to something different, or maybe even deleting the unused `this.entity`.

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
shwang:
```suggestion
                System.out.println("Warning skipped damage event - multiple damage events in one tick!");
                System.out.println("This damage event will be marked as stale=True.")
```

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @SubscribeEvent
+    public void onDeath(LivingDeathEvent event) {
+        if (event.getEntityLiving().equals(Minecraft.getMinecraft().player)) {
+            if (this.damageSource != null) {
+                // If the agent dies from damage, a LivingHurtEvent will precede the death event. Since death events do
+                // not include the damage amount we simply update the other fields to ensure the cause of death is
+                // properly reflected
+                this.damageSourceReplaced = true;
```
shwang, line 50: Looks like we are marking every "death from damage" event (and this will constitute a large fraction of our death events) as `stale=True` here -- is this intentional?

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
```diff
@@ -0,0 +1,142 @@
...
+    public void writeObservationsToJSON(JsonObject json, MissionInit missionInit) {
+        EntityPlayerSP player = Minecraft.getMinecraft().player;
+        JsonObject damage_json = new JsonObject();
+        json.addProperty("is_dead", player.isDead);
+        json.addProperty("living_death_event_fired", this.hasDied);
+
+        // Note to user that there could be inaccuracies here
+        if (this.damageSourceReplaced) {
+            damage_json.addProperty("stale", true);
+            this.damageSourceReplaced = false;
```
shwang: suggestion: Probably makes more sense to put this reinitialization assignment inside `resetDamage` along with the other reinitializations

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
shwang: > "we simply update the other fields to ensure the cause of death"

Doesn't look like you are updating other fields here -- instead you are reading from `this.damageSource` as usual?

shwang (approved):
Looks good -- I addressed PEP8 Speaks lint, and here are some optional suggestions

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
brandonhoughton: ahh great point

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
brandonhoughton: this.entity seems to only ever be the player - where as this.damageSource.getEntity() is the entity causing the damage - will update

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromDamageImplementation.java
brandonhoughton: Hmm I want to mark something regarding the fact that the damage event is overridden in the case that we can kill the agent without damaging it and want to know if that is the case?

brandonhoughton: I should be updating this.damageSource from event.getSource(); below

shwang: sounds good

PR #461, opened 2021-03-10
Revert large inventory for agents

brandonhoughton:
fixes openai/openai#6219
This implementation does not seem to work properly in the multi-agent setting, test with interactive mode before re-enabling


pzhokhov (approved):

brandonhoughton:
Closing this as @AdrienLE fixed what we believe is the big issue with large inventories (in that they are hard-coded in another place that we failed to override) and in consideration that additional mixins will now need to be disabled for the agent have vanilla sized inventory

PR #462, opened 2021-03-10
Adds respawning handler for agents - MUST UPDATE OLD ENV-SPECS

brandonhoughton:
Changes the default behavior of Minecraft to allow vanilla respawning (meaning player spawn rules are respected)
This does NOT run agentStartHandlers, so agent customization such as giving a random starting inventory will not persist after respawn but will be respected for each agent's first life

To keep the behavior from old experiments that end on death, clientQuitOnDeath and serverQuitOnDeath handlers have been added. These can be configured to end the episode for each agent after death, to end the episode for all agents when a single agent dies, and to end the episode once all agents die.

This PR is a WIP in the sense that the remaining experiments must have the serverQuitOnDeath(any_agent) specified to retain their original behavior.


brandonhoughton:
Formatting made ClientStateMachine.java messy - important changes are at 1684+

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Mixins/MixinMinecraftGameloop.java
```diff
@@ -268,14 +268,14 @@ private void runGameLoop() throws IOException
...
-    
+
     @Overwrite
     public void setIngameFocus()
     {
         if (!this.inGameHasFocus) {
             this.inGameHasFocus = true;
             this.displayGuiScreen((GuiScreen) null);
-            this.leftClickCounter = 10000;
+            this.leftClickCounter = 0;
```
brandonhoughton, line 278: If we leave this 10000 the agent has to wait that long before it is able to interact with anything again. Potentially a better fix would be to set this to 0 on respawn, but this should not affect the behavior of agents

minerl/Malmo/Schemas/MissionHandlers.xsd
```diff
@@ -1255,6 +1255,17 @@
...
     </xs:element>
 
+    <xs:element name="ServerQuitFromDeath">
+        <xs:annotation>
+            <xs:documentation>
+                Specifies a time limit that applies to all agents.
+            </xs:documentation>
+        </xs:annotation>
+        <xs:complexType>
+            <xs:attribute name="quitWhenAnyDead" type="xs:boolean" use="optional" default="false"/>
```
brandonhoughton, line 1265: Note the default behavior here is  different then the current default behavior for experiments (they are more simmilar to quitWhenAnyDead=true)

minerl/Malmo/Schemas/MissionHandlers.xsd
```diff
@@ -3109,6 +3120,15 @@
         </xs:complexType>
     </xs:element>
 
+    <xs:element name="AgentQuitFromDeath">
+        <xs:annotation>
+            <xs:documentation>
```
pzhokhov, line 3125: for agent to quit means to not respawn, correct?

brandonhoughton: Correct - they will stop being in the episode and the number of active agents will decrease unless the server also wants to quit when any agent dies

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/AgentQuitFromDeathImplementation.java
```diff
@@ -0,0 +1,76 @@
...
+
+public class AgentQuitFromDeathImplementation extends HandlerBase implements IWantToQuit
+{
+	private String quitCode = "";
+	private DamageSource deathSource = null;
+
+	
+	@Override
+	public boolean parseParameters(Object params)
+	{
```
pzhokhov, line 49: nitpick - your opening curly brackets seem inconsistent (sometimes they are on a new line, sometimes on the same line as function / class name

minerl/Malmo/Schemas/MissionHandlers.xsd
pzhokhov: what happens if AgentQuitFromDeath is true, <ServerQuitFromDeath quitWhenAnyDead=false>, and all agents die? Would the server restart in that case?

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/AgentQuitFromDeathImplementation.java
brandonhoughton: OK with either convention, I'll try and normalize. Off-topic should we lint all of the code base? I had a huge diff here because the code became auto-linted

minerl/Malmo/Schemas/MissionHandlers.xsd
```diff
@@ -1255,6 +1255,17 @@
         </xs:complexType>
     </xs:element>
 
+    <xs:element name="ServerQuitFromDeath">
```
pzhokhov, line 1258: maybe <ServerTimelimt>? QuitFromDeath seems misleading, if this is supposed to be a timelimit

minerl/Malmo/Schemas/MissionHandlers.xsd
brandonhoughton: The episode should end when there are 0 agents and the episode has started. I will check

minerl/Malmo/Schemas/MissionHandlers.xsd
brandonhoughton: Good catch

PR #463, opened 2021-03-11
Fix coal smelting test in monorepo

pzhokhov:


minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/CameraCommandsImplementation.java
```diff
@@ -99,9 +99,8 @@ public void onRenderTick(TickEvent.RenderTickEvent ev) {
                         this.currentYaw = player.rotationYaw;
                         this.currentPitch = player.rotationPitch;
                     }
-                    // TODO peterz: I am pretty sure this code is deprecated and does not match the docstring
-                    // i.e. no smooth panning is happening here
-                    // player.setPositionAndRotation(player.posX, player.posY, player.posZ, this.currentYaw, this.currentPitch);
```
JoostHuizinga, line 104: Maybe add a comment explaining the purpose of this code to prevent future users from commenting it out?

JoostHuizinga (approved):
Maybe add a comment explaining the purpose of this code to prevent future users from commenting it out?

Otherwise, merge whenever you see fit!

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/CameraCommandsImplementation.java
JoostHuizinga: Actually, just noticed, why did you change this.currentYaw into player.rotationYaw? Wasn't the point of this code to ensure that this.currentYaw is guaranteed to be equal to player.rotationYaw?

pzhokhov: no, not quite, it is to ensure client and server views of rotation angles (EntityPlayerSP.rotationYaw and EntityPlayerMP.rotationYaw, respectively) do not disagree. At least, this code was achieving such goal. For instance, in our coal smelting test failure, there were no camera actions involved; so disagreement between currentYaw and player.rotationYaw should not have affected it, however, it did.
The reason why I don't want to back to the code the way it was is because in lowLevelInput mode we inject mouse motions instead of using camera commands (to be able to move mouse in gui). Because of that, currentYaw and currentPitch are not updated, and the code as it was prevents camera movement.

PR #464, opened 2021-03-18
Fix at least some of the issues with large inventory

AdrienLE:
Basically ContainerPlayer also needs to be aware that there are 360 (or however many) items in the inventory, otherwise it doesn't get synced between the server and client.

### DIFF
--- a/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Mixins/MixinLargeInventorySlots.java
+++ b/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Mixins/MixinLargeInventorySlots.java
@@ -1,0 +1,52 @@
+package com.microsoft.Malmo.Mixins;
+
+import net.minecraft.entity.player.InventoryPlayer;
+import net.minecraft.inventory.Container;
+import net.minecraft.inventory.ContainerPlayer;
+import net.minecraft.inventory.Slot;
+import org.spongepowered.asm.mixin.Mixin;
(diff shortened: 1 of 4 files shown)


minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Mixins/MixinLargeInventorySlots.java
```diff
@@ -0,0 +1,52 @@
...
+                }
+            }
+        }
+        if (inventory == null) {
+            // TODO: should raise an error?
+            return;
+        }
+        while (missingIndices.size() > 0) {
+            int slot = missingIndices.first();
+            // Note: the first "slot" is the actual index, the other two are what the x and y position should be in the
```
brandonhoughton, line 44: Maybe we could be more clever with  slot % int(sqt(slot)), slot / int(sqt(slot))? It is possible to open the inventory as a human and would be nice to be able to see what was in it

brandonhoughton (approved):
Looks good to me - only comment applies to a future use case so feel free to ignore

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Mixins/MixinLargeInventorySlots.java
AdrienLE: My guess is that there are so many slots that no matter what it will look like a giant mess in the UI. I think we can wait until we cross that bridge for now.

PR #467, opened 2021-03-20
Don't update variants we don't actually care about

jietang:


jietang:
heh left this out of the prior commit

PR #468, opened 2021-03-24
Revert low_level mouse input code

brandonhoughton:


minerl/herobraine/hero/handlers/agent/actions/keyboard.py
```diff
@@ -61,7 +61,7 @@ def __init__(self, command, *keys):
         self.keys = keys
 
     def from_universal(self, x):
-        actions_mapped = set(str(k) for k in x['custom_action']['actions']['keys'])
+        actions_mapped = list(x['custom_action']['actions'].keys())
```
brandonhoughton, line 64: @pzhokhov Can you explain this change? Seems like we should keep set()

minerl/herobraine/hero/handlers/agent/start.py
```diff
@@ -80,29 +79,6 @@ def __init__(self, inventory : List[Dict[str, Union[str, int]]]):
         })
 
 
-class RandomInventoryAgentStart(InventoryAgentStart):
```
brandonhoughton, line 83: Perhaps this change should be kept too

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Mixins/MixinMinecraftGameloop.java
```diff
@@ -268,15 +264,4 @@ private void runGameLoop() throws IOException
 
         this.mcProfiler.endSection(); //root
     }
-    
-    @Overwrite
```
brandonhoughton, line 272: This change is needed for the re spawn code to work - will create conflicts. We should be sure to use this.leftClickCounter=0 to prevent the agent from getting stuck in the guiGameOver screen

minerl/herobraine/hero/handlers/agent/actions/keyboard.py
pzhokhov: tbh don't remember too well; there may have been an issue with repeated keys... The end result of human level input did not use this handler, so I suspect the change was not necessary.

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Mixins/MixinMinecraftGameloop.java
brandonhoughton: Fixed!

minerl/herobraine/hero/handlers/agent/actions/keyboard.py
brandonhoughton: Used the set notation for now

JoostHuizinga (approved):
Looks good to me, though I honestly don't have a clue what I am looking at for the most part.

PR #478, opened 2021-04-09
Env spec autoblacklist

shwang:
Adds `EnvSpec.auto_blacklist_demo` method and refactors publish.py blacklisting logic.

Summary of existing smoke test logic in `publish.py`:
 * Survival and Obfuscated EnvSpecs never blacklist.
 * "Obtain" environments auto-blacklist if reward == 1024 (excluding "SimonSays").
 * if forward action is never used, then auto-blacklist.
 * if attack action is never used, then auto-blacklist unless this is a "Navigate" demonstration.
 * if reward is 0, then auto-blacklist.

It seemed like the most straightforward way to refactor this logic in `publish.py` is to have `Obfuscated.auto_blacklist_demo` never blacklist (there is no analogous `EnvSpec` base class for "Survival" data right now), and the base class `EnvSpec` contain the rest of the logic. 

BASALT env specs will have an `auto_blacklist_demo` like this: https://gist.github.com/shwang/eb9bcf001d90836800d0c6f755b0b1cf

### DIFF
--- a/minerl/data/pipeline/publish.py
+++ b/minerl/data/pipeline/publish.py
@@ -352,19 +352,6 @@ def render_data(output_root, recording_dir, experiment_folder, black_list, lineN
                 raise e
 
 
-            # All Envs auto-blacklist by default if
-            # if action${attack,forward} exist and are never used.
-            # if reward is 0
-            #
-            #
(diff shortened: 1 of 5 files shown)


brandonhoughton (approved):

shwang:
(Ping Brandon when this stuff is tested)

PR #479, opened 2021-04-09
Fixed an issue where item placements would not be counted as item uses.

JoostHuizinga:
Fixed an issue in ForgeHooks.java where placing an item when the quantity of that item in your inventory is 1 would incorrectly result in the usage of "air" being increased, instead of the usage for that item.

### DIFF
--- a/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Mixins/MixinForgeHookStatsFix.java
+++ b/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Mixins/MixinForgeHookStatsFix.java
@@ -1,0 +1,114 @@
+package com.microsoft.Malmo.Mixins;
+
+import net.minecraft.block.state.IBlockState;
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.item.ItemBucket;
+import net.minecraft.item.ItemStack;
+import net.minecraft.nbt.NBTTagCompound;
+import net.minecraft.stats.StatList;
+import net.minecraft.util.EnumActionResult;
+import net.minecraft.util.EnumFacing;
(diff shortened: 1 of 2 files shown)


AdrienLE (approved):
Thanks

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Mixins/MixinForgeHookStatsFix.java
```diff
@@ -0,0 +1,114 @@
...
+                    }
+                }
+
+                // BUGFIX: Get the actual item of the stack, regardless of whether the stack is empty
+                if (itemstack.getCount() == 0) itemstack.setCount(1);
+                player.addStat(StatList.getObjectUseStats(itemstack.getItem()));
+                itemstack.setCount(newSize);
+                // BUGFIX: end
+                // Original code:
+                // player.addStat(StatList.getObjectUseStats(itemstack.getItem()));
```
brandonhoughton, line 106: If we were going to fully lean into mixins, this would be the perfect use case for the @inject annotation. Approving for now I will just leave a note to come back to this later

brandonhoughton (approved):

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/Mixins/MixinForgeHookStatsFix.java
JoostHuizinga: I have re-implemented the fix using a @Redirect, which seemed more appropriate than @Inject after looking into it.

brandonhoughton (approved):
Looks like an ideal fix to me! Item does not have a count and this should prevent air from being returned as the item in an empty item stack.

PR #480, opened 2021-04-09
BiomeGenerator: Lookup biome_id from str name

shwang:
Based on biome ids from version 1.11.2. Now `BiomeGenerator` can accept a string biome name in addition to the biome ID number.


minerl/herobraine/hero/mc.py
```diff
@@ -672,3 +672,71 @@ def strip_item_prefix(minecraft_name):
...
     return minecraft_name
+
+
+# Transcribed from: https://minecraft.gamepedia.com/index.php?title=Biome&oldid=1057817,
+#   which has the biomes as they were during the minecraft version at the time: 1.11.2
+# _m indicates that the terrain is more mountainous
+# _f indicates that the terrain has trees where the base biome does not.
+BIOMES_MAP = {
+    "ice_plains": 12,
+    "ice_plains_spikes": 140,
```
shwang, line 683: Check that biome map above 32 is not off by 2. (Brandon says that that above 32 is off by 2)

shwang: Seems legit!

image

brandonhoughton: Sweet! Looks spot on to me

brandonhoughton (approved):

PR #482, opened 2021-04-13
Add BASALT env specs, add rendering pipeline utilities, fix minerl.viewer

shwang:
Add BASALT envs (relies on #480 and #478).

Merges tests/basic_import_test.py into tests/env_smoke_test.py to test importing and stepping all environments, including BASALT envs.

Adds two new scripts to `minerl/data/pipeline/pipeline/`: `pipeline.py` and `pipeline_clean.py`. The first script runs all of the pipeline scripts in the canonical sequence (also can select subsets of scripts to run in sequence). The second script is a utility script that can selectively delete subsets of intermediate files produced by the rendering pipeline, including demonstration blacklists.

Adds interactive mode and menu to `download2.py`, which allows selective downloading of particular year/month/days of raw stream data.

Various changes to pipeline scripts so that they can generate required directories, download Minecraft, and compile parser.c on behalf of the user when applicable (some of this automatic tooling is only in `pipeline.py`.).

Fixes a few issues (DataPipeline failure to access `gym.spec()._kwargs["action_space"]` and DataPipeline exception when loading non-existent MonitorHandler data) that prevented minerl.viewer from int.

PR #488, opened 2021-04-29
Use replay mod that can take in nondefault resolution

shwang:
(**Merge procedure**: This branch and the displayed diff are built on top of #482. Merge that one first, and then switch PR base branch to dev to merge this PR into `dev`.)

Use the ReplayMod compiled from https://github.com/minerllabs/anvil/pull/25 to allow better render quality when environment variables are set.

I tested this with no environment variables (produced low res video) and with HD settings to produce https://youtu.be/FY2ULjl3Q-U


minerl/data/pipeline/render.py
```diff
@@ -245,8 +251,10 @@ def extract(fname):
                 assert not os.stat(action_mcbr).st_size == 0
 
                 # Run the actual parse action and make sure that its actually of length 0.
-                p = subprocess.Popen(["python3", "parse_action.py", os.path.abspath(
-                    action_mcbr)], cwd='action_rendering')
+                p = subprocess.Popen(
+                    ["python3", "parse_action.py", os.path.abspath(action_mcbr)],
+                    cwd='action_rendering',
+                )
```
brandonhoughton, line 259: Is this a linting only change? Also is there a way to open the python that is running instead of python3? This seems like it could break things if you had a funny python config

brandonhoughton (commented):
The readable diff looks good - obviously it is hard to see what the underlying mod is doing so I will review that first

minerl/data/pipeline/render.py
shwang: yeah, this is linting only.

shwang: Could try https://stackoverflow.com/a/2589722/1091722 to get path of same python?

brandonhoughton: Yeah something like `subprocess.Popen([sys.executable, "path/to/script.py"])` looks like it would work

shwang: Tested this out in the latest commit. It looks fine!

shwang:
Fresh `~/renderers/minecraft_*/` wasn't auto-generating the recording directory which `render.py` was expecting (leading to error), so I added a makedirs for that.

minerl/data/pipeline/render.py
brandonhoughton: Awesome thanks!

brandonhoughton (approved):

PR #494, opened 2021-05-07
pipeline: Render.py timeout if no writes in ten minutes

shwang:
Refactors repeated code in the render_videos function and adds a timeout to render videos -- a job will timeout if it hasn't written to an "artifact file" within ten minutes.

Also deletes debug.log at start of script (which, undeleted, was leading to broken state where every render job would end debug.log error line from previous job).

Changes polling behavior of render_videos to happen once per second, rather than as fast as possible.


minerl/data/pipeline/make_minecrafts.py
```diff
@@ -82,14 +87,15 @@ def main(n_minecrafts=NUM_MINECRAFTS):
             shutil.rmtree(target_mc_name)
         shutil.copytree(MINECRAFT_TEMPLATE, target_mc_name)
         xauth = os.path.join(target_mc_name, 'xauth')
+        launch_sh_content = launch_sh_template.format(
```
brandonhoughton, line 90: This change will not auto-close f, was that intentional?

shwang: I was just printing `launch_sh_content` earlier as a debug step -- this should leave `f`'s with-context unchanged at line 98

brandonhoughton: Ahh sorry! Missed the ordering of this!

brandonhoughton (approved):

PR #367, opened 2020-07-29
Fix data_pipeline.py memory issue

MichalOp:
Related to #356, should fix it for python 3.6. ~Workers still don't exit correctly in python >=3.7, but that doesn't cause _that significant_ memory leaks as far as I can tell.~

edit: 
Workers exit correctly now in python >= 3.7.
Closes #356. Closes #349.

PR #495, opened 2021-05-07
Batch Iter Cleanup

decodyng:
- Make it possible for `include_monitor` and `include_metadata` to be set to False and not error 
- Integrate changes from #367 which had previously only been merged into Master 
- Make processes into daemon processes so they terminate when the process closes

### DIFF
--- a/minerl/data/data_pipeline.py
+++ b/minerl/data/data_pipeline.py
@@ -397,12 +397,22 @@ class DataPipeline:
             Generator: A generator that yields (sarsd) batches
         """
         # Todo: Not implemented/
+        input_queue = multiprocessing.Queue()
+        trajectory_queue = multiprocessing.Queue(maxsize=preload_buffer_size)
+
+        workers = []
+        args = (input_queue, trajectory_queue)
+        for _ in range(self.number_of_workers):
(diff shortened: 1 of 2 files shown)


minerl/data/data_pipeline.py
```diff
@@ -391,38 +390,48 @@ def batch_iter(self,
             preload_buffer_size (int, optional): Increase to IMPROVE PERFORMANCE. The data iterator uses a queue to prevent blocking, the queue size is the number of trajectories to load into the buffer. Adjust based on memory constraints. Defaults to 32.
             seed (int, optional): [int]. NOT IMPLEMENTED Defaults to None.
             include_metadata (bool, optional): Include metadata on the source trajectory. Defaults to False.
-            include_monitor_data (bool, optional): Include monitor data (info dict) on the source trajectory. Defaults
+            include_monitor_data (bool, optional): Include monitor data (info dict) on the source trajectory. Defaults to False
```
brandonhoughton, line 393: Lol thanks for fixing the docstring

Miffyli:
Merged for now so branching off has working data-iter. No destructive changes apart from expected MineRL data version.

MishaLaskin:
Note: that meta data does not return even if you set `include_metadata=True`

PR #499, opened 2021-05-12
Add VillageSpawnDecorator

shwang:
New handler (and Malmo backend) for spawning player inside a village.

image

I tested and integrated these features -- not personally familiar with the implementation. Questions along those lines should go to @neel-alex.

### DIFF
--- a/minerl/Malmo/Schemas/MissionHandlers.xsd
+++ b/minerl/Malmo/Schemas/MissionHandlers.xsd
@@ -1241,6 +1241,18 @@
         </xs:complexType>
     </xs:element>
 
+    <xs:element name="VillageSpawnDecorator">
+        <xs:annotation>
+            <xs:documentation>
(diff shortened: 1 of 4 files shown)


minerl/Malmo/Schemas/MissionHandlers.xsd
```diff
@@ -1241,6 +1241,18 @@
         </xs:complexType>
     </xs:element>
 
+    <xs:element name="VillageSpawnDecorator">
+        <xs:annotation>
+            <xs:documentation>
+                Spawns the player in a village, assuming the spawn point is in a valid biome for a village (plains, savannah, taiga, desert). In the absence of a way to modify the agent spawn, also spawns the agent at (0, 0).
+            </xs:documentation>
+        </xs:annotation>
+        <xs:complexType>
```
shwang, line 1250: Try removing this

brandonhoughton (approved):

minerl/Malmo/Schemas/MissionHandlers.xsd
shwang: @brandonhoughton This was actually necessary! (The schema definition would fail if it were empty or something like that)

PR #509, opened 2021-06-04
herobraine..action: Fix deprecation collections warning

shwang:
```
minerl/herobraine/hero/handlers/agent/action.py:10
  /home/steven/PycharmProjects/minerl-3/minerl/herobraine/hero/handlers/agent/action.py:10:
DeprecationWarning: Using or importing the ABCs from 'collections'
instead of from 'collections.abc' is deprecated since Python 3.3,and in
3.9 it will stop working
    from collections import Iterable
```

### DIFF
--- a/minerl/env/_singleagent.py
+++ b/minerl/env/_singleagent.py
@@ -4,7 +4,6 @@
 
 from typing import Any, Dict, Tuple
 from minerl.env._multiagent import _MultiAgentEnv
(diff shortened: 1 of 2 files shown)


brandonhoughton:
Haha I just pushed something very similar =)

brandonhoughton (approved):
I moved the imports around so the follow the form, import, import x as, from y import x

PR #510, opened 2021-06-04
Fix and refactor Sphinx EnvSpec docs

shwang:
Refactors logic in Sphinx docs, minor linting, and fixes a couple of errors
using a refactored EnvSpec-to-Sphinx function.

- docs: remove trailing white-space
- docs: Fix NameError due to past-deprecated fn
- docs: Add EnvSpec-to-Sphinx utility and smoke test
- docs../environments: Fix broken env descriptions

PR #511, opened 2021-06-04
MVP Basalt env Sphinx docs

shwang:
Merge #510 first.

Some basic Basalt environment docs.

Includes a macro for generating starting inventory Sphinx docs inserts (This is important
for several BASALT tasks). Adds docs/README.md for doc generation and viewing quickstart.

Removes a few Sphinx build warnings by having the heading delimiters ("......") be the same
length as the headings.

- docs/README.md: Add brief README for quickstart
- Basic basalt docs


shwang:
Docs and TOC now looks like this:

image


We will need to find a better TOC hierarchy for diamond vs. basalt environments later.

shwang:
(Rebased onto `dev` after squash-merging #510)

brandonhoughton (approved):

Miffyli:
@shwang 

Small issue: Windows does not allow `:` in filenames (e.g. `docs/source/assets/basalt/waterfall8_0:30.gif` is invalid path). Can you update the filenames to use e.g. `-` instead?

shwang:
Sure — out of curiosity what happens to windows machine when it tries to
download this file? 😂

Miffyli:
She just throws her usual hissy-fit and pretends it is not there :') 

image

Edit: I guess this can be better than having some obscure, non-renderable control characters as filenames in Unix when you accidentally press one too many buttons, and then you are stuck with a file that requires Unix-Magic lvl 45 to access, let alone delete.

PR #514, opened 2021-06-08
Add Tag to Support BASALT vs Diamond Downloads

decodyng:


brandonhoughton (approved):
We use logger.error instead of an assert in other places, but this gets the point across so approving!

decodyng:
Apropos of a comment by @shwang yesterday, I added `competition` as a command line argument. One relevant thing here is that in the process I switched to using argparse, which I think is a bit cleaner to work with for multiple arguments, but this means that the syntax for calling a single experiment would switch from: 
`python -m minerl.data.download <exp>` to 
`python -m minerl.data.download --experiment <exp>` 

I think this is a good idea under the circumstances since, if you have the possibility to pass in both a `experiment` and `competition` flag, you need some way of distinguishing which you're passing in a given situation, but it does have the effect that any scripts people previously had that just directly called with the experiment command line would break. (I'm guessing a lot of people were just evoking the `download` command in Python rather than from the command line, though, since I don't think we ever referenced the latter in the docs?) 

@brandonhoughton Could you take another look given this change?

brandonhoughton:
The change looks good! But we need to update the downloading docs.

shwang:
Eliminate the all option (because we aren't uploading this dataset for now)

brandonhoughton (approved):

PR #515, opened 2021-06-09
Basalt docs 2

shwang:
Add action handler documentation for `use`. Expand documentation for `equip`.

Make BASALT gif filenames windows compatible.

Fix missing BuildVillageHouse docstring.

Add note to `minerl.viewer` Sphinx docs that mentions `use` and `equip` are not displayed for BASALT datasets.

Various expansions of BASALT environment docs.

Rename "Competition Environments" to "MineRL Diamond Competition Environments".

Rename "BASALT Environments" to "MineRL BASALT Competition Environments".


docs/source/environments/handlers.rst
```diff
@@ -107,23 +102,36 @@ Camera Control - :code:`camera`
...
+------------------------------------
 
 .. _equip:
 
-.. function:: equip : Enum('none', 'air', 'wooden_axe', 'wooden_pickaxe', 'stone_axe', 'stone_pickaxe', 'iron_axe', 'iron_pickaxe'))
+.. function:: equip : Enum('none', 'air', ..., 'other'))
 
     This action equips the first instance of the specified item from the agents inventory to the main hand if the
-    specified item is present, otherwise does nothing. :code:`none` is never a valid item and functions as a no-op
-    action. :code:`air` matches any empty slot in an agent's inventory and functions as an un-equip action.
```
brandonhoughton, line 120: What was the motivation for removing this line?

brandonhoughton: Ahh I see it is included in the docs below

docs/source/environments/index.rst
```diff
@@ -20,24 +20,26 @@ environment!
...
 
+
 Basic Environments
 =======================================
 
 .. warning::
 
-    The following basic environments are NOT part of the 2020 MineRL Competition! Feel free to use them for exploration
-    but agents may only be trained on `MineRL competition environments`_!
+    The following basic environments are NOT part of the 2021 MineRL Competition! Feel free to use them for exploration
```
brandonhoughton: Maybe any of the 2021 MineRL competitions to include basalt?

minerl/herobraine/env_specs/basalt_specs.py
```diff
@@ -422,24 +427,44 @@ def create_server_decorators(self) -> List[handlers.Handler]:
...
+Build a house in the style of the village without damaging the village. Give a tour of
+the house and then throw a snowball to end the episode.
+
+.. note::
+  In the observation and action spaces, the following (internal Minecraft) item IDs can be
+  interpreted as follows:
+
+    - ``log#0`` is oak logs.
+    - ``log#1`` is spruce logs.
+    - ``log2`` is acacia logs.
```
brandonhoughton: Maybe log2#0 here

brandonhoughton (approved):
Small nit picks but otherwise LGTM

docs/source/environments/index.rst
shwang: yeah, could change Competition! to `Competitions!` here.

I think the ~"You can't do this in Diamond (omitting any such rule for BASALT)" statement here is fine because as I understand it we are pretty lenient in BASALT with training methods (you can somehow train your agent on TreeChop if you wanted to?). @rohinmshah

shwang: I updated to the following. Let me know if I'm misunderstanding something related to the rules of either competition and we can adjust accordingly

image

brandonhoughton: Oh - good point perhaps then we should only say that these are not a part of MineRL Diamond

shwang:
Pretty sure this is failing because test_variant_utils.py is hanging probably because `log2#0` isn't recognized by Malmo . Will look into that tomorrow

shwang:
(based on local test)

docs/source/environments/index.rst
rohinmshah: Actually I don't think we should allow participants to train on TreeChop -- that's extracting an additional "did you chop a tree" reward that they shouldn't have access to.

(Doesn't matter very much for TreeChop, since they could make that reward themselves based on their inventory, but the general principle would be, don't use other environments since they might extract additional disallowed info from the Minecraft simulator.)

I might change it to:

> Feel free to use them for exploration but note that MineRL competition agents may only be trained on the corresponding competition environments.

brandonhoughton:
Hmm in addition it is timing out because with only 1 thread it takes  more than 20 minutes to eval =(

shwang:
It was hanging forever because Malmo didn't accept the MissionXML and was stuck repeating this message:

```
ERROR    minerl.env.malmo.instance.452f9a:malmo.py:756 [10:05:10] [EnvServerSocketHandler/INFO]: [STDOUT]: REPLYING WITH: MALMOERRORcvc-datatype-valid.
1.2.3: 'log2#0' is not a valid value of union type 'BlockOrItem'.                                                                                      
ERROR    minerl.env.malmo.instance.452f9a:malmo.py:756 [10:05:11] [EnvServerSocketHandler/INFO]: [STDOUT]: JAXB exception: javax.xml.bind.UnmarshalExce
ption                                                                                                                                                  
ERROR    minerl.env.malmo.instance.452f9a:malmo.py:756 [org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 1766; cvc-datatype-valid.1.2.3: 'lo
g2#0' is not a valid value of union type 'BlockOrItem'.]                                                                                               
ERROR    minerl.env.malmo.instance.452f9a:malmo.py:756 [10:05:11] [EnvServerSocketHandler/INFO]: [STDOUT]: REPLYING WITH: MALMOERRORcvc-datatype-valid.
1.2.3: 'log2#0' is not a valid value of union type 'BlockOrItem'.                                                                                      
ERROR    minerl.env.malmo.instance.452f9a:malmo.py:756 [10:05:12] [EnvServerSocketHandler/INFO]: [STDOUT]: JAXB exception: javax.xml.bind.UnmarshalExce
ption  
```

Fixed by https://github.com/minerllabs/minerl/pull/515/commits/ebc83cd0675ebea42a3dba87a7c16bb7842c0df4

docs/source/environments/index.rst
shwang: yeah, that sounds good @rohinmshah , I'll use that instead (I think that this wording covers the intent of the old message too :) )

brandonhoughton:
This is a good lesson that the way in which we handle Malmo errors is not ideal - in this case there is no way for us to upload a different XML so we should break rather than retry

PR #518, opened 2021-06-09
Fix Pyro4 exposes

Miffyli:
Bit of refactoring was needed to avoid trying to serialize sockets and other things. This code is only (mainly?) used on AICrowd servers, does not touch other functionality. Works on AICrowd evaluation servers.


brandonhoughton (approved):
Looks good to me

PR #519, opened 2021-06-11
Don't print so much on each equip

brandonhoughton:


Miffyli (approved):
LGTM! Should help with log sizes a bunch :)

PR #520, opened 2021-06-11
Don't log so many seconds in the count down!

brandonhoughton:


Miffyli (approved):
LGTM! Any reason why everything was logged in the past? For completeness of logs?

decodyng (commented):
I'm in a similar boat as @Miffyli in terms of not having a clear sense of the motivations for the original code, so, looks good to me, but with low context behind that

brandonhoughton:
This was just the malmo default - the only real use-case would be for human demonstrations so that the player knows the time remaining, or if someone wanted to add the chat observation to their agent.

PR #521, opened 2021-06-11
Speed up list comprehention

brandonhoughton:
The previous method was significantly slower than any other handler (no_op call accounted for > 2.5% of the env.step) 
Here we try to speed this up by replacing no_op with default dict


minerl/herobraine/hero/handlers/agent/observations/inventory.py
```diff
@@ -79,11 +82,11 @@ def from_hero(self, obs):
         :param obs:
         :return:
         """
-        item_dict = self.space.no_op()
+        item_dict = defaultdict(str)
```
shwang: Probably want to use `defaultdict(int)` here to default to zero instead of ""

shwang:
Looks like Dict returns an OrderedDict here (different from unordered defaultdict, in the unlikely case that breaks any assumptions):

https://github.com/minerllabs/minerl/blob/725424df4371504c4a0f3b64d08d3fec75102d0e/minerl/herobraine/hero/spaces.py#L316-L318

minerl/herobraine/hero/handlers/agent/observations/inventory.py
shwang:
```suggestion
        item_dict = {item_id: 0 for item_id in self.item_list}  # Much faster than Dict.no_op()
```

Should be just as fast, and more explicit for the user/developers.

shwang (changes requested):
The main slowdown in calling `Dict.no_op` is probably from processing recursive `no_ops` for all of the numpy arrays `item_id: Discrete(###)` -- which becomes `item_id: 0` in all cases anyways.

Since we know that the no_op space is simply a `item_id: 0` in this case we can just do a dictionary comprehension which should be just as fast as the `defaultdict` approach, and more straightforward for users/developers (having a defaultdict embedded inside the observation dict can lead to confusion).

minerl/herobraine/hero/handlers/agent/observations/inventory.py
brandonhoughton: Good point - I've updated

minerl/herobraine/hero/handlers/agent/observations/inventory.py
```diff
@@ -92,14 +95,13 @@ def from_hero(self, obs):
             key = util.get_unique_matching_item_list_id(self.items, type_name, stack['metadata'])
             assert stack["quantity"] >= 0
             assert stack["metadata"] in range(16)
-            if key is not None:
+            if key is not None and self.space.contains(key):
```
brandonhoughton: @shwang Can you confirm that with item variant squashing this is the correct behavior?

brandonhoughton:
Notably, this PR changes the default behavior of the item observation space. Previously if there were items in the inventory not in your item_list those would still be returned - now they are compacted into the other catagory if other is included in items else ignored. 
Also this reverts the from_universal change that does not count empty slots properly which @shwang may have better insight into if that is an issue or not

minerl/herobraine/hero/handlers/agent/observations/inventory.py
```diff
@@ -93,35 +96,29 @@ def from_hero(self, obs):
...
-                    # instead of dicts where `slot["name"] == "air"`.
-                    continue
-                item_type = mc.strip_item_prefix(stack['name'])
-
-                if item_type == "air" and "air" in item_dict:
-                    item_dict["air"] += 1
+                item_type = mc.strip_item_prefix(stack['name']) if len(stack.keys()) != 0 else "air"
+
+                if item_type == "air" and item_type in self.items:
+                    item_dict["air"] += 1  # This lets us count empty slots -non default MC behavior
```
shwang: The change to counting empty (i.e. air) slots seems fine to me -- the reason the tests are failing right now is maybe because they are asserting different "air" counts now that the counting logic is changed.

Could you look into that (using the test logs)?

https://github.com/minerllabs/minerl/blob/e8d16a790c2915a10aa6f36b03c74668704a5dc9/tests/unit/test_inventory_obs.py#L148-L156

shwang: If there is something confusing about these tests, let me know and I can check them on my local machine

minerl/herobraine/hero/handlers/agent/observations/inventory.py
```diff
@@ -93,35 +96,29 @@ def from_hero(self, obs):
             assert stack["quantity"] >= 0
             assert stack["metadata"] in range(16)
             if key is not None:
+                if key not in self.items:
+                    key = self._other
```
shwang: if key isn't None, then `key in self.items` is guaranteed by  `key = util.get_unique_matching_item_list_id(self.items, type_name, stack['metadata'])`.

minerl/herobraine/hero/handlers/agent/observations/inventory.py
```diff
@@ -93,35 +96,29 @@ def from_hero(self, obs):
...
                 item_dict[key] += stack["quantity"]
```
shwang:
```suggestion
            if key is not None:
                item_dict[key] += stack["quantity"]
            elif self._other in self.items:
               item_dict[self._other] += stack["quantity"]
```

shwang: "other" isn't in any of the FlatInventoryObservation item lists for any of our EnvSpecs (include ObtainDiamond), show probably should check before inserting this into the returned item_dict.

brandonhoughton: You mean check that this works? It will only be added if self._other is in self.items

minerl/herobraine/hero/handlers/agent/observations/inventory.py
shwang: The `self.space.contains(key)` should be guaranteed by the `key is not None`. Maybe make the `self.space.contains` into assert statement (if we keep it at all)?

minerl/herobraine/hero/handlers/agent/observations/inventory.py
shwang: Ah, by "check" I meant add the `elif self._other` check.

minerl/herobraine/hero/handlers/agent/observations/inventory.py
brandonhoughton: The tests fail because we check if the obs.sample is in the space - but we expected a numpy object (with attribute .shape) where I was returning an int. I should have this fixed by returning the default dict of np.array(0)

minerl/herobraine/hero/handlers/agent/observations/inventory.py
brandonhoughton: OK I adjusted test_inventory_obs.py to reflect the correct behavior

minerl/herobraine/hero/handlers/agent/observations/inventory.py
brandonhoughton: Good point - I've adjusted the logic such that we return self._other iff key = None and self._other in self.items

shwang (approved):

PR #524, opened 2021-06-14
Fix SyntaxWarnings from use of 'is'/'is not' for literal checks

mrahtz:
When trying to use MineRL with Python 3.9, I get:

```
>>> import minerl
minerl/minerl/data/data_pipeline.py:137: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  x[0] if x[0] is not 'pov' else 'z')
minerl/minerl/herobraine/hero/handlers/agent/reward.py:212: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  not self.fired[bl['type']] or bl['behaviour'] is not "onlyOnce"):
minerl/minerl/herobraine/env_specs/navigate_specs.py:146: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if top is "normal":
```

This PR replaces `is` with `==` and `is not` with `!=`.

I've tested by installing `minerl` from my branch, and confirming that I can `import minerl` without issue.

Thanks!

### DIFF
--- a/minerl/data/data_pipeline.py
+++ b/minerl/data/data_pipeline.py
@@ -134,7 +134,7 @@ class DataPipeline:
         x = list(target_space.spaces.items())
         target_space.spaces = collections.OrderedDict(
             sorted(x, key=lambda x:
-            x[0] if x[0] is not 'pov' else 'z')
+            x[0] if x[0] != 'pov' else 'z')
         )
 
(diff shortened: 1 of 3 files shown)


brandonhoughton (approved):
Thanks for the PR, looks good to me!

PR #525, opened 2021-06-17
Refactor data_ordering test and reintroduce to CI

shwang:
Refactors `data_ordering_test.py` for clearer error logs
(now logs the set of observation or action keys are mismatched
between the dataset and the observation/action space).

Also runs these tests on CI and on every environment with a dataset
(lists of Basalt competition envspecs, diamond envspec, and envspecs with
datasets are also added to the codebase in this PR)

```
- docs: Move competition env spec list to `envs`
- data_ordering_test: Refactor for clearer errors
- data_ordering_test: Reintroduce _check_space test
- data_ordering_test: comment/function rewording
- data_ordering_test: Remove repeat mark
```


shwang:
@brandonhoughton: I uploaded the "minimal"  random subset of the newly rendered dataset to `s3://minerl/v3/data_texture_0_low_res_minimal.tar`. You should be able to use this to update the dataset used by the testing server

shwang:
(Waiting on @brandonhoughton to upload dataset to Azure)

brandonhoughton:
Data has been uploaded and the path corrected!

brandonhoughton:
Currently tests hang - not sure what the cause of that is but they did not complete in over 2 hours

shwang:
I confirmed that this branch passes on my local dataset (which I am uploading to S3).

Not clear whether this works with the data I had previously posted on S3, oops! (Leads to error where BASALT env name diverged with dataset name, but definitely not hanging on my local machine).

brandonhoughton:
Cool - ping me when the data is uploaded and I can get it updated on azure

shwang:
sounds good, I'll most likely reupload another (at least minimal) dataset tonight using #539

shwang:
@brandonhoughton New dataset is uploaded

shwang:
We shouldn't expect this to work with the v4 dataset right now (since I copied diamond's v3 data, with its conflicting observation space, into the v4 dataset)

PR #531, opened 2021-06-18
malmo metadata: Mod "in-the-wild" metadata by 16

shwang:
Sometimes the metadata value we see in Malmo is greater than 15, leading to RuntimeExceptions. The reason that this is happening is because "in-the-wild" item stacks, like arbitrary blocks that the player picks up, could have non-variant metadata inside it.

This PR hopefully prevents these exceptions by modding "in-the-wild" ItemStack.getMetadata() by 16. 

- malmo metadata: Mod "in-the-wild" metadata by 16.
- equip handlers: Add new metadata assertions

### DIFF
--- a/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromEquippedItemImplementation.java
+++ b/minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromEquippedItemImplementation.java
@@ -43,7 +43,7 @@ public class ObservationFromEquippedItemImplementation extends HandlerBase imple
             {
                 String type = MineRLTypeHelper.getItemType(itemToAdd.getItem());
                 jobj.addProperty("type", type);
-                jobj.addProperty("metadata", itemToAdd.getMetadata());
+                jobj.addProperty("metadata", itemToAdd.getMetadata() % 16);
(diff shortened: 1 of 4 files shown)


shwang:
@brandonhoughton: Just to make sure I'm not going crazy want to double-check that the variant bits are indeed always the lowest 4 bits 😂

brandonhoughton:
I thought we had discussed that metadata can go to 2^(16) because it is a short? Should we still mod things this way - it will make pickaxes in variant aware cases look like they are changing item type

shwang:
My intention is to let stuff like pickaxe overloaded damage stuff slide —
this should be fine because the python-side code shouldn’t be looking at
its metadata anyways.

My worry here is with other items, whose variant we care about, that could
have data other than variant stored with them (not sure if this actually
can happen in Minecraft, to be fair)

Right now, due to my sanity check raising a RuntimeError, if anything in
the inventory has a metadata/damage greater than 15, then the entire
inventory handler (or sometimes equip handler) will fail. So it’s important
to have some fix (other potential fixes are dropping the sanity assertion,
which seems risky, or figuring out programmatically which items actually
have metadata fields).

Also, really since what we care about is just the variant bits here, it
would have been better if we named this field “variant” instead, but oh
well (one reason I think we went with metadata here is because the XML
schema already uses another field called variant which is an Enum).

brandonhoughton:
OK - this is a good hot-fix until we decide how to deal with variant for items that have changing meta-data properties!

PR #533, opened 2021-06-22
Ignore non metadata from items without sub-types

brandonhoughton:


brandonhoughton:
Currently this PR is broken by the use of filled maps which will eventually throw assertion errors because the metadata value will increase beyond 16. This is intended however as maps with distinct metadata values ARE distinct - we should simply remove the assert statements once we've validated other items are working as indended

brandonhoughton:
Banners are another item with subtypes worth investigating

brandonhoughton:
another  thought here - there are different use-cases for item matching, specifically for equip item we want to ignore the metadata of the item equipped (e.g. match any pick_axe) but for maps and planks we only want to equip the item specified or else you get a random map or plank variant from your inventory. For giving agents inventory, one would imagine that we would actually like to be able to give agents a damaged pick_axe, such as when reusing a previous inventory. I think this is ok to address later I just wanted to bring this up

minerl/Malmo/Minecraft/src/main/java/com/microsoft/Malmo/MissionHandlers/ObservationFromEquippedItemImplementation.java
```diff
@@ -43,7 +43,11 @@ public static JsonObject getInventoryJson(@Nullable ItemStack itemToAdd){
             {
                 String type = MineRLTypeHelper.getItemType(itemToAdd.getItem());
                 jobj.addProperty("type", type);
-                jobj.addProperty("metadata", itemToAdd.getMetadata() % 16);
+		if (itemToAdd.getHasSubtypes()){
```
shwang: Oh neat -- you can actually check if there are subtypes here!

shwang:
We're busy fixing several installation errors in the BASALT baselines package -- can't look deeply into this right now, but I can get back to it later if that's helpful

shwang (approved):
LGTM

PR #535, opened 2021-06-24
Buffered Batch Iterator + Tests

decodyng:
(Mostly copied from `minerl_dev` Slack message)

A prototype of a (more approximately, though not perfectly) IID replacement/alternative for our current batch iterator

Basically, it works by:

- Maintaining a buffer of (approximately) some fixed size
- Starting out by filling that buffer with <size> timesteps
- At each point in the generator, it returns a batch_size batch, and updates the buffer if the space remaining in the buffer is greater than the average length of trajectories we’ve seen so far (definitely a heuristic, but it doesn’t seem like this needs to be perfect, just ballpark, since trajectories don’t tend to be more than one order of magnitude different in length)
- When it runs out of trajectories, it just runs through the remainder of the buffer, before optionally re-setting the buffer and the “available trajectories” set again if we’re running for multiple epochs
- This doesn’t give us perfectly all timesteps, but AFAICT the error is just that we drop any remaining timesteps at the end if we end up with remainder < batch_size, to avoid returning a truncated batch at the end of the epoch, so for most reasonable batch sizes this loss should be negligible

This PR also includes changes to test this new iterator in `test_batch_iter` and reference it (and describe the known bug with existing `batch_iter`) in the documentation

### DIFF
--- a/docs/source/tutorials/data_sampling.rst
+++ b/docs/source/tutorials/data_sampling.rst
@@ -40,17 +40,27 @@ Or we can simply download a single experiment
 For a complete list of published experiments, `checkout the environment documentation`_. You can also download the data
 in your python scripts 
 
-Now we can build the datast for :code:`MineRLObtainDiamond-v0`
+Now we can build the dataset for :code:`MineRLObtainDiamond-v0`
+
+There are two ways of sampling from the MineRL dataset: the deprecated but still supported `batch_iter`, and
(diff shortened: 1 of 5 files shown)


brandonhoughton (approved):

brandonhoughton:
Looks good to me including the additional warning. Reminder that we should update the docs when this gets merged

decodyng:
> Looks good to me including the additional warning. Reminder that we should update the docs when this gets merged

By "update the docs" do you mean re-build the docs, or do you mean actually write additional docs beyond what's in this PR?

PR #538, opened 2021-06-30
basalt envs: Fix and test snowball throw wrapper

shwang:
Old code failed to throw snowball because of off-by-one error -- the equipped_item type observation would no longer be "snowball" on the same tick that `actions["use"] == 1`, because the observation we receive in `observation, reward, done, info = super().step(action)` comes from the step after the snowball is thrown. Therefore, we used to never terminate the episode on an actual snowball throw because `observation` here would report that no item was equipped (the snowball was already thrown by the time we see `observation`).

### DIFF
--- a/minerl/herobraine/env_specs/basalt_specs.py
+++ b/minerl/herobraine/env_specs/basalt_specs.py
@@ -92,20 +92,23 @@ class EndAfterSnowballThrowWrapper(gym.Wrapper):
     def reset(self):
         self._steps_till_done = None
         obs = super().reset()
+        snowball_count = obs["inventory"]["snowball"]
+        assert snowball_count == 1, "Should start with a snowball"
         return obs
 
     def step(self, action: dict):
         observation, reward, done, info = super().step(action)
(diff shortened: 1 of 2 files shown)


shwang:
Instead of aligning the `use` action and the equipped item observation, I'm going to look at the snowball count instead from now on, and assert that there is exactly one snowball in the inventory upon episode start.


```
(basalt) steven@spinach ~/PycharmProjects/basalt_competition_baseline_submissions % python scrap/parse_npzs.py
throw tick: 957          estimated throw time: 47.85 sec
episode duration: ~180.0 sec
2643
SUMMARIZING TEN TICKS AROUND SNOWBALL THROW TIME:
num snowball in inventory [1, 1, 1, 1, 1, 0, 0, 0, 0, 0]
equipped item type ['snowball', 'snowball', 'snowball', 'none', 'none', 'none', 'none', 'none', 'none', 'none']
```

Note that inventory empirically updates two steps later than equipped item type for some reason, so this means that the episode will end two ticks later than expected (12 ticks total).

decodyng (approved):
lgtm!

PR #539, opened 2021-07-01
publish: Fix empty-inventory on air-stack bug

shwang:
Fix an error where empty stacks errored on "KeyError: variant" when "air" was not an item inside `item_list`.

Also, reduces verbosity of `publish.py` stage of rerendering pipeline by only outputting "unknown item" messages when `MINERL_DEBUG_LOG` env var is set.

### DIFF
--- a/minerl/herobraine/hero/handlers/agent/observations/inventory.py
+++ b/minerl/herobraine/hero/handlers/agent/observations/inventory.py
@@ -114,8 +114,13 @@ class FlatInventoryObservation(TranslationHandler):
             for stack in slots:
                 item_type = mc.strip_item_prefix(stack['name']) if len(stack.keys()) != 0 else "air"
 
-                if item_type == "air" and item_type in self.items:
-                    item_dict[item_type] += 1  # This lets us count empty slots -non default MC behavior
+                if item_type == "air":
+                    # We don't expect to actually see any slots with `stack['name'] == 'air'`,
+                    # (air should be an empty dictionary stack) but if we do,
+                    # `stack['name'] == 'air'` is also covered under this case.
+                    if item_type in self.items:
+                        # This lets us count empty slots non-default MC behavior
+                        item_dict[item_type] += 1
                 else:
                     unique = util.get_unique_matching_item_list_id(
                         self.items, item_type, stack['variant'])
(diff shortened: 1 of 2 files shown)


minerl/herobraine/hero/handlers/agent/observations/inventory.py
```diff
@@ -114,8 +114,13 @@ def from_universal(self, obs):
...
 
-                if item_type == "air" and item_type in self.items:
-                    item_dict[item_type] += 1  # This lets us count empty slots -non default MC behavior
+                if item_type == "air":
+                    # We don't expect to actually see any slots with `stack['name'] == 'air'`,
+                    # (air should be an empty dictionary stack) but if we do,
+                    # `stack['name'] == 'air'` is also covered under this case.
+                    if item_type in self.items:
+                        # This lets us count empty slots non-default MC behavior
+                        item_dict[item_type] += 1
```
brandonhoughton, line 123: Maybe I am misreading this but I can't find the difference here?

shwang: Oops, could have been more clear about this. It's pretty subtle -- if "air" isn't in `self.items`, then in the old regime we go to the `else:` case, which errors because `item_type == "air"` implies that `stack == {}`.

shwang:
(note-to-self -- rerender upon merge to dev)

minerl/herobraine/hero/handlers/agent/observations/inventory.py
brandonhoughton: Ahh so it is just a debugging change?

shwang: No, in the old regime we error out due to keyerror.

When we enter the else case (because `"air" not in self.items`, then we error out on `stack["variant"]` because `stack == {}`. Moving the `"air" in self.items` conditional into a nested `if` statement in this patch prevents us from ever entering the `else:` clause whenever we have an empty `stack`.

shwang: (This change fixes the KeyError "variant" problem)

brandonhoughton: Ahh I missed the indentation implication

brandonhoughton (approved):

PR #541, opened 2021-07-01
Hotfix: Robustify Village spawn procedure

shwang:
Previously we had an issue where players would die suddenly after `reset()` (within the first 3 frames) because they spawned inside a wall.

This would lead to videos where we see all of the players items flying in every direction, including the snowball, which would trigger an episode end due to the EndEpisodeOnSnowballThrowWrapper. (It's not clear why the snowball ends the episode, and not the actual player death).

This hotfix combines two hacks to reduce the chance of player death.
(1) Spawn the player 2 blocks above the suggested village spawn location. Empirically, this seems to reduce the probability of dying immediately on reset. We can reduce the probability of dying even further by spawning the player 6 blocks above the suggested village spawn location.

(2) Add a wrapper that upon `env.reset()`, checks that the player hasn't dropped all their items for 30 noop actions. If the player hasn't died in that time frame, then the spawn is likely OK. If the player died in that time frame, then retry reset up to 3 times.

### DIFF
--- a/minerl/herobraine/wrappers/retry_reset_on_early_death_wrapper.py
+++ b/minerl/herobraine/wrappers/retry_reset_on_early_death_wrapper.py
@@ -1,0 +1,69 @@
+import logging
+
+import gym
+
+
+def _agent_seems_alive(obs: dict) -> bool:
+    """Return True if there are any items in inventory (a telltale sign of agent death is
+    dropping all items)"""
+    inventory = obs["inventory"]
+    for count in inventory.values():
+        if count > 0:
+            return True
+    return False
+
+
+class RetryResetOnEarlyDeathWrapper(gym.Wrapper):
+    """
+    Apply this wrapper directly over a MineRLEnv with a VillageSpawnDecorator to
+    reset() when the agent dies ~instantly due to spawning inside a wall.
+
+    Can also make it seem as if the agent spawned on the ground instead of in the air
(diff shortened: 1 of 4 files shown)


minerl/herobraine/wrappers/retry_reset_on_early_death_wrapper.py
```diff
@@ -0,0 +1,69 @@
...
+class RetryResetOnEarlyDeathWrapper(gym.Wrapper):
+    """
+    Apply this wrapper directly over a MineRLEnv with a VillageSpawnDecorator to
+    reset() when the agent dies ~instantly due to spawning inside a wall.
+
+    Can also make it seem as if the agent spawned on the ground instead of in the air
+    because as a side effect this wrapper will call `self.env.step(noop_act)` several
+    times to check if the player has died immediately after `reset()`.
+
+    WARNING: This wrapper may have unexpected interactions with `seed()`, which otherwise
```
decodyng, line 25: Yeah, this is the thing I'm most uncertain about vis a vis this solution. One thing we had talked about with @rohinmshah  and @Miffyli was whether we should just select seeds that we're _pretty sure_ don't have those problem for our own evaluation (though having this fixed in general is obviously still good for participants running their own tests)

decodyng (approved):
This looks good to me, with one general comment about seeding strategy

PR #544, opened 2021-07-06
pipeline script: Add regex flags for processing particular demonstrations

shwang:


brandonhoughton (approved):

setup.py
```diff
@@ -153,7 +153,7 @@ def build_minecraft(source_dir, build_dir):
 
 
 setuptools.setup(
-    name='minerl',
+    name='minerl-shwang',
```
shwang:
```suggestion
    name='minerl',
```

PR #545, opened 2021-07-07
Add links to projects/papers that use MineRL

Miffyli:
Review from a native speaker would be welcome before merging :)


brandonhoughton (approved):
Looks awesome!

PR #551, opened 2021-07-09
download: Update docs and assertions

shwang:
Updates `download` function documentation, in particular noting the `experiment` => `environment` change (sorry for this stealth change earlier), and noting which arguments are "internal" on which ones are external.

I also moved the important (public-facing) arguments in `download()` earlier in the argument list, and added some assertions

Broader question:
Does it seem like the right idea to document the `download` function as a low-level function, and point people towards using the `python -m minerl.data.download` script instead?

### DIFF
--- a/minerl/data/download.py
+++ b/minerl/data/download.py
@@ -17,30 +17,73 @@ import logging
 from minerl.data.version import DATA_VERSION, assert_version
 from minerl.herobraine import envs
 import tempfile
+from typing import Optional
 import coloredlogs
 
 logger = logging.getLogger(__name__)
 
 
-def download(directory=None, resolution='low', texture_pack=0,
-             update_environment_variables=True, disable_cache=False,
-             environment=None, competition=None):
-    """Downloads MineRLv0 to specified directory. If directory is None, attempts to 
-    download to $MINERL_DATA_ROOT. Raises ValueError if both are undefined.
-    
+def download(
+        environment: Optional[str] = None,
+        competition: Optional[str] = None,
+        directory: Optional[str] = None,
+        resolution: str = 'low',
+        texture_pack: int = 0,
+        update_environment_variables: bool = True,
+        disable_cache: bool = False,
+) -> None:
+    """Low-level interface for downloading MineRL dataset.
+
+    Using the `python -m minerl.data.download` CLI script is preferred because it performs
+    more input validation and hides internal-use arguments.
+
+    Run this command with `environment=None` and `competition=None` to download a minimal
+    dataset with 2 demonstrations from each environment.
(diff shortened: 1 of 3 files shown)


minerl/data/download.py
```diff
@@ -17,30 +17,73 @@
...
 logger = logging.getLogger(__name__)
 
 
-def download(directory=None, resolution='low', texture_pack=0,
-             update_environment_variables=True, disable_cache=False,
-             environment=None, competition=None):
-    """Downloads MineRLv0 to specified directory. If directory is None, attempts to 
-    download to $MINERL_DATA_ROOT. Raises ValueError if both are undefined.
-    
+def download(
```
KarolisRam, line 26: Is changing the order of the variables worth it here? It would change the functionality of `download('data')`, like in the examples here:
https://minerl.readthedocs.io/en/latest/tutorials/index.html?highlight=download(#installation

KarolisRam (commented):
looks good, just not 100% sure about changing the order of variables.

minerl/data/download.py
shwang: Oh good find -- that would be a confusing mismatch

shwang: What about we put directory first?

It does seem useful to push the unused/internal arguments to the end.

shwang: Here's the mismatching part which we will have to edit regardless of the argument order because `experiment` is no longer an argument:

image

Maybe the quickest way to fix this would just be removing this section from the "Installation" docs? There's already a ["Downloading and Sampling the Dataset" section](https://minerl.readthedocs.io/en/latest/tutorials/data_sampling.html) right underneath it in the user guide, and downloading the dataset isn't really part of installing MineRL.

KarolisRam: Directory first would probably be safe enough, yes.

shwang: ok, I moved the directory arg to the front, and I removed the outdated docs I mentioned in the comment above.

minerl/data/download.py
```diff
@@ -17,30 +17,73 @@
...
+        update_environment_variables: For internal use only. If True, then export of
+            MINERL_DATA_ROOT environment variable (note: for some os this is only for the
+            current shell).
+        disable_cache: If False (default), then the tar download and other temporary
+            download files are saved inside `directory`.
+
+            If disable_cache is False on
+            a future call to this function and temporary download files are detected, then
+            the download is resumed from previous download progress. If disable_cache is
+            False on a future call to this function and the completed tar file is
```
brandonhoughton, line 76: Should this second clause be True?

brandonhoughton (approved):
Lgtm - see comment on potential typo

PR #570, opened 2021-07-27
data_pipeline: Better FileNotExistsError for Windows

shwang:
Earlier, DataPipeline would give a very uninformative error message on Windows machines when looking for a nonexistent trajectory.


Miffyli (approved):
LGTM (after local Windows tests)!

PR #571, opened 2021-07-30
download: Add warning msgs about minimal dataset

shwang:


Miffyli (approved):
LGTM! Was going to add an explanation on what is the minimal dataset, but this already is explained in the data docs.

PR #575, opened 2021-08-04
Dockerfile: Add -dev and -data-processing stages

shwang:
The -dev stage includes an install of the latest minerl `dev` commit (upon
Docker image build). The -data-processing stage adds packages,
environment variables, and volumes used for processing raw demonstration
data into datasets.

Each stage confirmed to work on my computer. The Dockerfile comments contains instructions on where to find these images on DockerHub and example commands for using the data-processing container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,21 @@
-FROM ubuntu:18.04
+FROM ubuntu:18.04 as circleci-base
 
-# This Docker image, tagged and uploaded as `springulum/minerl-circleci-base`,
+# The `circleci-base` stage Docker image, tagged and uploaded as `springulum/minerl-circleci-base`,
 # bundles all of the Ubuntu packages needed to run and test minerl locally.
 # It does not include the minerl repository because this is meant to be downloaded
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -49,3 +50,73 @@ RUN apt-get install -y --no-install-recommends \
...
+#      python3 -m minerl.data.pipeline.pipeline -j 8
+#
+# - Show script help:
+# docker run -it --rm springulum/minerl-data-processing python3 -m minerl.data.pipeline.pipeline --help
+
+# Prevent interactive `apt-get install` again, as described above.
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update -q && apt-get install -y --no-install-recommends \
+    p7zip-full \
```
shwang, line 102: Seems to be missing apt install x11vnc leading to "silent" render failures

shwang (commented):
asdf

shwang:
I'm gonna merge this in now so that we have these dockerfiles available in case we want to start data processing again.

Since this is just a change to this Dockerfile, it should not affect anyone's work.

PR #576, opened 2021-08-09
Add tasks to docs

Miffyli:
Add the workshop tasks to docs along with link to baselines (they serve as nice examples).

### DIFF
--- a/docs/source/api/env.rst
+++ b/docs/source/api/env.rst
@@ -12,8 +12,8 @@ simulator Malmo which enables **synchronous**, **stable**, and
 MineRLEnv
 ===========
 
-.. automodule:: minerl.env.core
-    :members: MineRLEnv
+.. automodule:: minerl.env._multiagent
+    :members: _MultiAgentEnv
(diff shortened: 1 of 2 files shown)


docs/source/api/env.rst
```diff
@@ -12,8 +12,8 @@ simulator Malmo which enables **synchronous**, **stable**, and
 MineRLEnv
 ===========
 
-.. automodule:: minerl.env.core
-    :members: MineRLEnv
+.. automodule:: minerl.env._multiagent
```
brandonhoughton: There is also _singleagent env - do we want Multi-agent support to be the default here?

Miffyli: Right, I was worried that it would not have enough docs, but generating it it seems to offer description of the basic step/reset funcs, so that should be enough :D

Miffyli:
Made the change. If ok, feel free to merge right away :)

brandonhoughton (approved):

PR #587, opened 2021-09-11
Add info on xvfb alternative

Miffyli:
Add docs on the virtual display trick Tencent researchers kindly shared.


brandonhoughton (approved):
Looks good to me!

Issue #512, opened 2021-06-07
Metadata does not return with batch_iter even if include_metadata=True


MishaLaskin:
```
dat = minerl.data.make('MineRLTreechopVectorObf-v0')
for struct in tqdm(dat.batch_iter(batch_size=16, seq_len=32, 
                                  num_epochs=1, preload_buffer_size=32,
                                  include_metadata=True)):
    print(struct)
```
struct outputs without any metadata in this example

Miffyli:
Hey. I can take look at this tomorrow. Probably a silly logic typo somewhere in the PR.

Related PR #495

Miffyli:
After bit of digging I realized I did not actually push all changes I made to #495 (I felt like I missed something). I will be continuing tomorrow on a fix for this. Thanks for pointing this out!

Miffyli:
I removed `include_metadata` keyword for `batch_iter` as proper support for it proved to be very tricky to implement. I recommend using `load_data` if you wish to get the metadata info as well.

For future, see [`fix/batch_iter_2`](https://github.com/minerllabs/minerl/tree/fix/batch_iter_2) for WIP progress.

PR #588, opened 2021-09-14
Update data_pipeline.py

### LINKED ISSUE #572: Bug in data_pipeline
Hi,
    There is a bug in current version(0.4.0) minerl about include_metadata. The following is a link to the position of bug.

    https://github.com/minerllabs/minerl/blob/dev/minerl/data/data_pipeline.py#L129

    If we want to get infor like success, we need to enable include_metadata. But in current version, this is controled by include include_monitor_data, which I think is a typo.

Regards,
Shenghan

AntonioAlgaida:
closes #572

Propose change in lane 129 to obtain "meta" when load_data.

### DIFF
--- a/minerl/data/data_pipeline.py
+++ b/minerl/data/data_pipeline.py
@@ -126,7 +126,7 @@ class DataPipeline:
         remainder = iter(seq[5:])
 
         monitor_seq = next(remainder) if include_monitor_data else None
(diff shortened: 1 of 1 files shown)


Miffyli (approved):
LGTM! Thanks for fixing this silly mistake.

(PS: You might want to check #512 as well)

PR #591, opened 2021-09-22
Hotpatch missing rewards from end of episode

Miffyli:
Last reward of "ObtainX" specs was not included because the `onPickup` or other triggers of `RewardForPossessingItemImplementation` were not triggered (the QuitOnPosession handler was too quick killing things?).

This hotpatch delays the quitting by several `doIWantToQuit` calls, which seems to fix the issue.


danijar:
Not sure if this is the same issue, but we noticed this with the MineRL 0.4.4 ObtainDiamond task, described in [DreamerV3](https://arxiv.org/pdf/2301.04104v2) appendix. The episode ends once a diamond block is broken, which makes the item pop out, but there often isn't enough time for the item to hit the player to be added to the inventory. Sometimes, the item also lands on the floor, but the player doesn't have time to step on it to collect the item and the reward.

Miffyli:
@danijar Good catch, could be the same issue. People before Dreamer did not reach diamonds in v0.4 or older, so that code-path was never really executed 😆 . It probably is bit different situation, though, since it uses the Malmö library to implement episode ends and reward picking. I thought it should work fine, but haven't tried it. 

I do not have time to debug/fix things though, but if people find fixes and do PRs, happy to review them when I have time!

PR #611, opened 2022-01-15
Chat Action Handler!

trigaten:
This allows the execution of Minecraft chat commands.
Take note that the Handler and the Text MineRLSpace behind the handler aren't fully implemented (warnings are thrown).

### DIFF
--- a/minerl/Malmo/Minecraft/build.gradle
+++ b/minerl/Malmo/Minecraft/build.gradle
@@ -28,7 +28,7 @@ buildscript {
         compileClasspath {
             resolutionStrategy.eachDependency { DependencyResolveDetails details ->
                 if (details.requested.group == 'org.apache.logging.log4j') {
-                    details.useVersion '2.15.0'
+                    details.useVersion '2.16.0'
                     details.because 'zero-day-exploits on log4j <2.15.0'
                 }
             }
--- a/setup.py
+++ b/setup.py
@@ -168,7 +168,7 @@ setuptools.setup(
     name='minerl',
     # TODO(shwang): Load from minerl.version.VERSION or something so we don't have to update
     # multiple version strings.
-    version=os.environ.get('MINERL_BUILD_VERSION', '0.4.3'),
+    version=os.environ.get('MINERL_BUILD_VERSION', '0.4.4'),
     description='MineRL environment and data loader for reinforcement learning from human demonstration in Minecraft',
     long_description=markdown,
     long_description_content_type="text/markdown",


minerl/herobraine/hero/handlers/agent/actions/chat.py
```diff
@@ -0,0 +1,46 @@
+__author__ = "Sander Schulhoff"
+__email__ = "sanderschulhoff@gmail.com"
+
+from minerl.herobraine.hero.handlers.agent.action import Action
+import minerl.herobraine.hero.spaces as spaces
+
+# TODO add more command support (things like what commands are allowed) from Malmo
+# TODO ensure both agents can send chats, not just first agent (again, check Malmo)
```
Miffyli: "Both" should probably mean "other"? :)

Also could this be related to Minecraft op-ness where only first agent has op powers (i.e. they can call commands like `tp`)?

docs/source/api/herobraine.rst
```diff
@@ -123,6 +123,14 @@ Smelt
     :undoc-members:
     :show-inheritance:
 
+Chat
```
Miffyli: An update to main documentation/examples would be nice, because otherwise this feature can be missed easily :).

Maybe even a new subsection under `Hello World, you first agent` tutorial.

minerl/env/_multiagent.py
```diff
@@ -270,6 +273,11 @@ def step(self, actions) -> Tuple[dict, dict, bool, dict]:
         if not self.done:
             assert STEP_OPTIONS == 0 or STEP_OPTIONS == 2
 
+            # add chat action if there is one
+            if self.next_chat_message:
+                actions["agent_0"]["chat"] = self.next_chat_message
```
Miffyli: Bit of confusion here, but does one have to add the Chat handler to environments to be able to use this? If so, add note about this to documentation and/or throw appropriate error here that informs about this.

Ideally I would like to see that all environments have this option on by default, but I guess that'd involve modifying all existing environments (I could do this) :)

Miffyli (commented):
Some quick major notes, will test the code once I get back to a linux machine :)

minerl/herobraine/hero/handlers/agent/actions/chat.py
trigaten: yep, changed "Both" to "other"

Interesting idea. I don't believe that to be the case.

trigaten:
> Some quick major notes, will test the code once I get back to a linux machine :)

Cool, I have tested it by adding a couple lines to the MLGWB gym and the added feature worked.

minerl/env/_multiagent.py
trigaten: Yes, if it isn't enabled _some_ error should get thrown. I will look into doing that.

You should only need to modify `SimpleEmbodimentEnvSpec`.

Miffyli: Adding doc/tutorial on how to do this would be enough for now (lets leave modifying other envs for later).

docs/source/api/herobraine.rst
trigaten: Did one better, a new tutorial :)

minerl/env/_multiagent.py
trigaten: Did one better, a new tutorial :)

PR #622, opened 2022-03-27
Fix interactor

magic-sword:
* Sleep 10 sec immediately after sending a message to avoid  "Invalid session id" on _multiagent.py
* Fixed socket to send hello messages on interactor.__main__.request_interactor

I tried to fix Interactive Mode which is broken in version 0.4

If you run env.make_interactive (), the socket times out after the following message is displayed:
"Could not authorize you against Realms server: Invalid session id"
This could be avoided by sleeved for a few seconds (8 seconds or more) after sending the message.

I didn't know the detailed mechanism to solve it.
Please let me know if there is a better solution.

Fixed the socket where the message of interactor.request_interactor is sent.
The following function refers to the instance instead of the socket as an argument, so it needs to be modified.
"_MultiAgentEnv._TO_MOVE_hello (sock)"

With the above, the error can be corrected and it can be confirmed that the player has joined the environment of the agent.
However, the Minecraft launcher freezes without displaying the game screen.

I tried to find a solution, but I couldn't figure it out with my help as a beginner.
If anyone knows the cause, would you please tell me?

### DIFF
--- a/docs/source/tutorials/minerl_tools.rst
+++ b/docs/source/tutorials/minerl_tools.rst
@@ -75,8 +75,9 @@ Interactive Mode :code:`minerl.interactor`
 
 .. warning::
 
-    The interactor does not function in MineRL version v0.4. If you wish to use this utility,
-    install an older version of MineRL ``pip install minerl=0.3.7``.
+    Interactor works in MineRL versions 0.3.7 and 0.4.4 (or above). 
+    Install 0.3.7 with ``pip install minerl==0.3.7``, or the newest MineRL
+    with ``pip install git+https://github.com/minerllabs/minerl.git@dev``.
 
 
 Once you have started training agents, the next step is getting them to interact with human players.
--- a/minerl/env/_multiagent.py
+++ b/minerl/env/_multiagent.py
@@ -478,6 +478,7 @@ class _MultiAgentEnv(gym.Env):
             # the port/ip of the master agent send the remaining XMLS.
 
             self._send_mission(self.instances[0], agent_xmls[0], self._get_token(0, ep_uid))  # Master
+            time.sleep(10)
             if self.task.agent_count > 1:
                 mc_server_ip, mc_server_port = self._TO_MOVE_find_ip_and_port(self.instances[0],
                                                                               self._get_token(1, ep_uid))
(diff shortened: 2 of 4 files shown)


Miffyli:
Hey! Thanks for the PR! This is a simple solution, but if it works, I am happy to merge it as an immediate fix :). I will test this out tomorrow or day after.

I might update the `self.shape` change, once I test if it would work without a change on gym==0.19. You can also remove the commented out code for `comms` thing, no need to leave older code in comments :)

magic-sword:
Thanks for the reply.
I am impressed that you responded so quickly.

Fixed a problem with PEP8.
I also removed the old code that was commented out.

Thank you for your kind cooperation.

Miffyli:
I tested this out and it works flawlessly on few attempts I tried! I updated the doc to reflect that interactor will work in repo code.

I was under the impression something failed in your end? Can you clarify when your code failed? 

Btw, since this code combines Python + Java + obscure server/client communication, I would say debugging it is not an easy task and you should feel proud for figuring these things out (even if the fix seems crude) :). Out of personal curiosity, how did you debug or figure out the solutions for this?

magic-sword:
I'm glad it worked!

My development environment is incomplete.
The Docker environment isn't available, I'm just running the cloned code on Windows.
Therefore, the reason it doesn't work in my environment may be due to an incomplete environment.

The work in the complete environment was unknown, so I was relieved to hear that it worked.


To be sure, debugging was difficult because the VS Code tools didn't work.
The pdb module barely worked, so I stepped through it line by line for debugging.

It is a by-product of the miracle that was born in this process that I discovered that pausing seems to work.

It's a small fix, but I'm very proud to be involved in some interesting interactivity.

I would like to thank you for publishing this wonderful project for easy interactive AI research and for your quick response.



I look forward to the day when this fix will be merged and you can try out the interactivity with the latest code!

Miffyli (approved):
Ahh gotcha! Indeed dockering and all that does interfere with interactive mode :).

Hehe sounds like any good old debugging of first doing hard work and then randomly stumbling to solution ^^. Hope you have learned a thing or two in the process! 

Merging this. Once again: thank you for your time and contributions! :). This was a very good PR from a newcomer!

PR #623, opened 2022-03-29
Add link to second place winner for MineRL 2020

adzcai:
I believe I found the paper for the second place winner -- I tracked it down from the video of the Round 2 finalists. I didn't participate in the competition, but I'm fairly confident this is the right link.


Miffyli:
LGTM! Thanks for spotting this! I try to keep up with arxiv digests but every now and things slip past me 😅 . I updated the page to reflect current 2020 leaderboard (there were retroactive changes).

Issue #621, opened 2022-03-21
Failed to start an environment


Kamilkime:
I'm just starting with MineRL - I've downloaded the package, and tried using the example code:
```python
import gym
import minerl
import logging

logging.basicConfig(level=logging.DEBUG)
env = gym.make('MineRLNavigateDense-v0')

obs = env.reset()
```

However - I'm not able to start the envitonment, I instantly get gradle build errors: https://pastebin.com/TJJfkKnk
Java is setup correctly - all environment variables point to JDK 8 folder. I use Python 3.9 on Windows 10: https://pastebin.com/JQKEpWKF
I have these packages installed for Python: https://pastebin.com/TFETXbDG

Any ideas what might be wrong here? I work with Minecraft and Java almost every day... But here I'm totally out of ideas :(

Miffyli:
Hmm that is a new one :o . I tried to dig up if anybody else has had the same issue but could not find related issues. My wild guess is that it is something with using OpenJDK on Windows. On linux this works, but on Windows people (me including) have used the Oracle JDK which works like a charm. Any possibility to try that out?

Also: note that you have to install the package with `pip install minerl`. Downloading the source from github and running directly won't work, and I recall having issues doing `pip install git+[github page]` on Windows, too.

Kamilkime:
> Also: note that you have to install the package with `pip install minerl`. Downloading the source from github and running directly won't work, and I recall having issues doing `pip install git+[github page]` on Windows, too.

That's how I did it, I followed the documentation for installation and first steps - except for Java installation, that one I already had done.

> My wild guess is that it is something with using OpenJDK on Windows. On linux this works, but on Windows people (me including) have used the Oracle JDK which works like a charm. Any possibility to try that out?

After dealing with their stupid downloads and accounts - I've installed Oracle JDK... But it changed nothing. I tried reinstalling minerl - still nothing. And then I tried running that python code from CLI with admin privileges - and it finally worked. Earlier I've installed minerl from admin CLI - maybe that was the problem? Also fun fact - that version of gradle (although working with java 8) does not support lambdas in groovy, my init script broke everything for a second (https://pastebin.com/Yj0HDe34).

And also I have one other question - while running the agent the window looks like this:
![obraz](https://user-images.githubusercontent.com/10383567/159386470-8c50b3c6-255e-46ab-bd97-8f7ebc76aa6c.png)


Should it be just empty like that? Or is there a way to see what the client is actually doing?

[...]

PR #625, opened 2022-04-02
Fix interactor on Windows os

magic-sword:
This is a fix to fix an issue where the player's MineRL window does not update on Windows OS.
This will also update the MineRL window on Windows OS when you run the python -m minerl.interactor 6666 command.
Details of the problem can be found [here](https://github.com/minerllabs/minerl/issues/621).

* Added display update option to launchClient.bat (-updateDisplay)
* When executing interactor module, added display update option to launchClient execution command

This fix is volatile and has the following concerns:
* Delayed reflection of settings
The option received by launchClient.bat is reflected in the MineRL window by writing it to the "run \ config \ malmomodCLIENT.cfg" file.
However, the reflection of the config file is delayed for some reason, causing the following problems.

1. Execute `env.reset ()` in the agent environment
→ The agent's MineRL window opens and is not updated (assumed behavior)

2. Run the `minerl.interactor` module
→ MineRL window for player opens and does not update (problem behavior)

3. Run the `minerl.interactor` module again
→ The MineRL window for the player opens and is updated (assumed behavior)

4. End the execution of the agent and player

5. Execute `env.reset ()` in the agent environment
→ The agent's MineRL window opens and is updated (problem behavior)

6. Execute `env.reset () `again in the agent environment.
→ The agent's MineRL window opens and is not updated (assumed behavior)

* Impact on Linux OS
Added new options to the launchClient command.
This may cause an error in launchClient.sh running on the Linux OS.
It looks like I need to modify launchClient.sh, but I'm not familiar with sh files and I don't have a Linux test environment.
If you have a Linux OS, please test for any errors.

 If you have a better idea or review, please let me know.
Yours sincerely

### DIFF
--- a/minerl/Malmo/Minecraft/launchClient.bat
+++ b/minerl/Malmo/Minecraft/launchClient.bat
@@ -28,7 +28,7 @@ setlocal enableDelayedExpansion
 ::
 :: The order of the definitions is not important.
 ::
-set "options=-port:0 -replaceable: -scorepolicy:0 -env: -runDir:run  -performanceDir:NONE -seed:NONE"
+set "options=-port:0 -replaceable: -scorepolicy:0 -env: -runDir:run  -performanceDir:NONE -seed:NONE -updateDisplay:"
 
 :: Set the default option values
 for %%O in (%options%) do for /f "tokens=1,* delims=:" %%A in ("%%O") do set "%%A=%%~B"
@@ -89,11 +89,15 @@ if "!-env!"=="true" (
     echo B:env=!-env!
     echo }
     )
+
+echo runtype {
 if "!-replaceable!"=="true" (
-    echo runtype {
     echo B:replaceable=!-replaceable!
-    echo }
     )
+if "!-updateDisplay!"=="true" (
+    echo B:updateDisplay=!-updateDisplay!
+    )
+echo }
 ) > "run\config\malmomodCLIENT.cfg"
 
 :launchLoop
--- a/minerl/Malmo/Minecraft/launchClient.bat
+++ b/minerl/Malmo/Minecraft/launchClient.bat
@@ -28,7 +28,7 @@ setlocal enableDelayedExpansion
 ::
 :: The order of the definitions is not important.
 ::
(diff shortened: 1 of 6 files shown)


Miffyli:
Thanks for your throughout work and details on potential error situations! You have really thought this all out :)
(I also want to apologize for saying earlier that things work on Windows for me -- I was mistaken. I was testing under Linux, and indeed things fail on Windows side).

Unfortunately, as you said, this is bit tricky solution that you have to launch interactor twice (and then afterwards the MineRL window had the "flashing problem"). While interactor is a nifty utility we should have working all around, this may cause more headaches than it may fix :(.

It has been a while since I fiddled with MineRL code in depth, but maybe you could somehow deduce that Minecraft is being ran in interactive mode (inside Java code) and thus enable the window-refreshing, without using the config file? I.e. see what enabling interactor mode changes inside Java, and based on that enable the window refreshing.

A small heads up: we might update MineRL code substantially as we intend to re-do the MineRL BASALT competition. Depending on the exact changes, it may overwrite your contributions. Just a bit of a warning so sudden re-work does not come up as a surprise ^^

magic-sword:
Thank you for a better idea!

I compared the agent and the player to see the difference between the two when the display starts up.
As a result, it turned out that only the agent seems to receive the mission init and initialize the display.

Taking advantage of this difference, I decided to set the display update flag to false at some initialization of the mission init.

As a result, the problem was completely solved in my trial!

The fix is ​​also very small, just a few lines in VideoHook.java and TimeHelper.java.
I think the negative effects on the Linux OS have disappeared.

However, when I open the inventory on the player's display, the item seems to look strange.
This is another issue, so I'll take a closer look at another opportunity.
image

Thank you for letting me know the notes.
I understand that the fix can be overwritten.
If it is overwritten and broken, fix it again and make a pull request  :)

If you like, please review or merge

Thank you for your precious time.

Miffyli:
LGTM! :) 

Thanks for the change! Now it indeed works beautifully and should not interfere with anything :). Thanks again for spending your time looking into this!

As for game UI: I think it is somewhat wonky due to all Malmo mods, and fixing it is likely a large chore, so I would leave it as is for now :) (it was also never meant to be played by humans, anyway. For interactive mode you should use chat commands to do things).

Merging this.

Miffyli (approved):

PR #613, opened 2022-01-24
ChatAction handler

Miffyli:
Originally made by @trigaten in #611 but I managed to nuke it completely beyond any recognition.

@trigaten are you ok if I merge this version or would you like to make new PR to get the credits and all that?

PR #638, opened 2022-06-27
Added chat action (and commands) in v1.0.0

Phylliida:
This change modifies `EnvServer.java` to accept a "chat" commad, and adds chat.py [from the dev branch](https://github.com/Phylliida/minerl/blob/dev/minerl/herobraine/hero/handlers/agent/actions/chat.py) so we have `H.ChatAction()`. It also enables commands (cheats).

Implemented because I've been watching the OpenAI agents and sometimes I want to help them out when they get stuck.

To use this, just put

```
actions['chat'] = 'thing to say'
```

and for example, modify `minerl\herobraine\env_specs\human_controls.py` `HumanControlEnvSpec` create_actionables to have

```
] + [H.CameraAction()] + [H.ChatAction()]
```

instead of

```
] + [H.CameraAction()]
```

on line 70. Note if you are using the VideoPreTraining you'll also need to modify the code in VideoPreTraining `agent.py` `validate_env` (I just comment it out) and then you'll be able to run commands and chat.

Because this also enables cheats you can do something like

```
actions['chat'] = '/give @p diamond 3'
```

To give the agent 3 diamond.

For example, in order to call actions in VideoPreTraining `run_agent.py` I just did this hacky thing

```
        if os.path.isfile("command.txt"):
          f = open("command.txt", "r")
          minerl_action['chat'] = f.read()
          f.close()
          os.remove("command.txt")
```

inserted after this

```
    while True:
        minerl_action = agent.get_action(obs)
```

then I can put my command in command.txt and then copy it into the running directory. There's definitely a better way to do this (like hosting a http server) but this works good enough if you just want to be watching AI lets plays.

By the way, if you want to enable peaceful mode to give the agents a easier time, just modify line 2241 of `scripts\mcp_patch.diff`, this line:

```
+        WorldSettings worldSettings = new WorldSettings(worldName, GameType.SURVIVAL, false, Difficulty.HARD, true, new GameRules(), DatapackCodec.VANILLA_CODEC);
```

to

```
+        WorldSettings worldSettings = new WorldSettings(worldName, GameType.SURVIVAL, false, Difficulty.PEACEFUL, true, new GameRules(), DatapackCodec.VANILLA_CODEC);
```

### DIFF
--- a/scripts/mcp_patch.diff
+++ b/scripts/mcp_patch.diff
@@ -1818,7 +1818,7 @@ diff -u1 -rN MCP-Reborn-vanilla/src/main/java/com/minerl/multiagent/RandomHelper
 diff -u1 -rN MCP-Reborn-vanilla/src/main/java/com/minerl/multiagent/env/EnvServer.java MCP-Reborn/src/main/java/com/minerl/multiagent/env/EnvServer.java
 --- MCP-Reborn-vanilla/src/main/java/com/minerl/multiagent/env/EnvServer.java	1970-01-01 01:00:00.000000000 +0100
 +++ MCP-Reborn/src/main/java/com/minerl/multiagent/env/EnvServer.java	2022-06-23 00:28:31.911515382 +0100
-@@ -0,0 +1,687 @@
+@@ -0,0 +1,699 @@
 +package com.minerl.multiagent.env;
 +
 +import com.google.common.base.Charsets;
@@ -2238,7 +2238,7 @@ diff -u1 -rN MCP-Reborn-vanilla/src/main/java/com/minerl/multiagent/env/EnvServe
 +        }
 +
 +
-+        WorldSettings worldSettings = new WorldSettings(worldName, GameType.SURVIVAL, false, Difficulty.HARD, false, new GameRules(), DatapackCodec.VANILLA_CODEC);
++        WorldSettings worldSettings = new WorldSettings(worldName, GameType.SURVIVAL, false, Difficulty.HARD, true, new GameRules(), DatapackCodec.VANILLA_CODEC);
 +        DimensionGeneratorSettings dms = DimensionGeneratorSettings.fromDynamicRegistries(DynamicRegistries.getImpl(), seed, generateFeatures, bonusChest);
 +        mc.createWorld(worldName, worldSettings, DynamicRegistries.getImpl(), dms);
 +    }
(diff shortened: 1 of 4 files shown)


scripts/mcp_patch.diff
```diff
@@ -2238,7 +2238,7 @@ diff -u1 -rN MCP-Reborn-vanilla/src/main/java/com/minerl/multiagent/env/EnvServe
 +        }
 +
 +
-+        WorldSettings worldSettings = new WorldSettings(worldName, GameType.SURVIVAL, false, Difficulty.HARD, false, new GameRules(), DatapackCodec.VANILLA_CODEC);
++        WorldSettings worldSettings = new WorldSettings(worldName, GameType.SURVIVAL, false, Difficulty.HARD, true, new GameRules(), DatapackCodec.VANILLA_CODEC);
```
Miffyli, line 2241: I assume this change enables cheats/commands on the server?

Phylliida: yup

Miffyli:
Hey! Thanks for this amazing (and quick!) PR :). Great to see people contributing back like this!

I will hold off the merge on this for now: this command would go against the BASALT competition rules, and we do not want to give the wrong impression to participants by providing tools they are not supposed to use the competition. We could have two branches, but that too could be confusing. I am open for suggestions on how to handle this situation, though.

(sidenote, but check out the `tests/test_human_interactor.py` :) . This too might be useful for poking agent to the right direction.)

scripts/mcp_patch.diff
```diff
@@ -2390,17 +2390,29 @@ diff -u1 -rN MCP-Reborn-vanilla/src/main/java/com/minerl/multiagent/env/EnvServe
...
++           }
++        }
++    }
 +
 +    private static KeyboardListener.State constructKeyboardState(String actions) {
 +        List<String> keysPressed = new ArrayList<>();
 +        for (String action: actions.split("\n")) {
 +           String[] splitAction = action.trim().split(" ");
-+           if (!splitAction[0].equals("camera") && !splitAction[0].equals("dwheel")) {
++           if (!splitAction[0].equals("camera") && !splitAction[0].equals("dwheel") && !splitAction[0].equals("chat")) {
```
Phylliida, line 2415: this and the check below are needed to prevent parseInt from throwing an exception when given a non int string and crashing the server

Miffyli:
Oh, one more small request: if you could add a minimal "test" script (not actual unit-test, just a file like other tests under `tests` which do a minimal interaction loop to demonstrate something), that would be nice :)

jeasinema:
Thanks for this patch! @Phylliida I think you may also want to modify [this file](https://github.com/minerllabs/minerl/blob/dev/minerl/herobraine/hero/spaces.py) to allow sampling from the new action space with chat command? [ref](https://github.com/minerllabs/minerl/pull/613/files#diff-2df8a47049915765c18557841bb05853c9f6f08057531ef693c0ccbb6b4e3965)

Miffyli (approved):
LGTM! Sorry for taking this long to look into this. I will merge this once competition is over :)

Issue #594, opened 2021-10-10
Installation Error


Ryul0rd:
Followed official installation instruction and got this error.
[minerl_logs.txt](https://github.com/minerllabs/minerl/files/7318386/minerl_logs.txt)

Miffyli:
Hey. You seem to have  a wrong version of java installed, based on all the errors. Make sure you have JDK 8 installed (other versions will not work. You can check this with `javac -version`)

Ryul0rd:
It says I've got the following: javac 1.8.0_292
Is this not correct?

Miffyli:
Oh, that should be correct o: . I was assuming java version was at fault given all the compilation errors.
I just tried doing fresh install of MineRL with same JDK version, and things worked out. Things to try out:

- Make sure that the correct javac version is used for installation (if you are running installation in some code editor, for example, it might change Java path).
- Try clearing caches (gradle and pypi caches) and reinstalling MineRL (e.g. removing `.gradle` directory in your home directory).
- If you are on Windows, check the [Windows FAQ](https://minerl.readthedocs.io/en/latest/notes/windows.html)

Ryul0rd:
Found the problem. I had installed Java 8 correctly but my system was still using the default Java 11 install. All I had to do was manually switch to the new installation using `sudo update-alternatives --config java`. Might want to consider making a note of this in the official installation instructions as this seems like it would be a somewhat common issue.

djsamseng:
> Found the problem. I had installed Java 8 correctly but my system was still using the default Java 11 install. All I had to do was manually switch to the new installation using `sudo update-alternatives --config java`. Might want to consider making a note of this in the official installation instructions as this seems like it would be a somewhat common issue.

Thank you! Also want to second adding this to the installation docs as I'm sure many developers would have newer versions of java installed

PR #639, opened 2022-06-27
Add installation instructions for verifying JDK 8 on linux

djsamseng:
This came up for a few of us on [this Github issue](https://github.com/minerllabs/minerl/issues/594#issuecomment-939578146). Let me know if this helps!


Miffyli (approved):
Thanks for the effort of doing the PR! I clarified the language a bit, but other than it was LGTM :)

PR #641, opened 2022-06-29
Make headless=True run xvbf

neverix:
It doesn't do anything anyway and there's no way to turn it on so why not

### DIFF
--- a/minerl/env/malmo.py
+++ b/minerl/env/malmo.py
@@ -581,7 +581,7 @@ class MinecraftInstance(object):
             cmd += ['-maxMem', self._max_mem]
 
         if headless:
-            cmd = ["/usr/bin/xvfb-run", "-a"] + cmd
+            cmd = ["xvfb-run", "-a"] + cmd
 
         cmd_to_print = cmd[:] if not self._seed else cmd[:-2]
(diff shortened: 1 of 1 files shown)


Miffyli:
Hey! Thank you for this: this actually was a thing on my todo list but deep in the "good to haves" list 😅 . I will definitely test this out. Just a small request: change from full/absolute path to just `xvfb-run` command, as sometimes people might the commands available somewhere else, and then they can always add things to PATH if `xvfb-run` is not in PATH by default.

neverix:
> test this out

It is completely untested, I'm still struggling to install MineRL locally

> change from full/absolute path 

True, I wrote that thinking psutil couldn't read from PATH

Miffyli:
Cheers, I will merge this once it is fully fleshed out (e.g., you can pass the "headless" argument from gym.make all down here or so), so that people can use it. In current form it not set-able, but that should not be a big work  :)

Btw please first raise issues or discuss changes on Discord before making PRs, and only make PRs when you are confident it could be merged. This is to reduce spam for everyone included.

neverix:
Yeah, this was kind of a joke when I first made the PR. I did not expect this to actually be useful. Will note for future contributions

PR #642, opened 2022-06-29
Make VPT work out of the box

neverix:
Or is this for the 64x64 models?


Miffyli (approved):
Cheers, it indeed was some old default parameters that should probably not be there (easier for user) :)

PR #644, opened 2022-07-04
v0.3.7 readthedocs fix

shwang:
Add .readthedocs.yml that uses pip install . instead of pip install -r requirements.txt.
Add READTHEDOCS toggle to avoid installing java dependence.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,14 @@
+version: 2
+
+sphinx:
+    configuration: docs/source/conf.py
+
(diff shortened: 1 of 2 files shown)


shwang:
Looks like this branch is tied to PYPI somehow (based on the name). I can merge this in but not sure if that would cause something to go out of sync.

Miffyli:
Nah it is good to merge; the name was just something I came up with when cleaning up 0.3.7 for things 😅 .

PR #664, opened 2022-09-03
fix life status not reading from info problem

xieleo5:
I've tried the lifeStats observation, and found that it is actually not changing at all. I took a look at the code and found that the `'life_stats'` key is missing. The info provided by java side is like
```
{
  'life_stats': {
    'xx': xx,
    'xx': xx
  }
}
```
If we don't add `'life_stats'` before `hero_keys`, when we do `walk_dict`, it will directly call `info['xx']` but not `info['life_stats']['xx']`. Then it will return the `default_is_missing` value.

### DIFF
--- a/minerl/herobraine/hero/handlers/agent/observations/lifestats.py
+++ b/minerl/herobraine/hero/handlers/agent/observations/lifestats.py
@@ -40,7 +40,7 @@ class LifeStatsObservation(KeymapTranslationHandler):
     def __init__(self, hero_keys, univ_keys, space, default_if_missing=None):
         self.hero_keys = hero_keys
         self.univ_keys = univ_keys
-        super().__init__(hero_keys=hero_keys, univ_keys=['life_stats'] + univ_keys, space=space,
+        super().__init__(hero_keys=['life_stats'] + hero_keys, univ_keys=['life_stats'] + univ_keys, space=space,
                          default_if_missing=default_if_missing)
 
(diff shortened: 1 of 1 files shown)


Miffyli:
Hey! Thanks for the PR. I will try to look into this tomorrow. Meanwhile, one thing that confuses me is the `univ_keys` parameter which is set as `life_stats`. Would you have idea what is the purpose of that argument? (PS: this code is inherited from other devs and I have not touched this part myself).

xieleo5:
I guess the `hero_key` intended to get value from its parent class, which means the parent class need to first unpack one layer of the info, then pass the unpacked value to all its children. While the `univ_key` is just unpack the value from the origin info dict passed from the java side. So I think maybe the better solution to fix this is to add a step to unpack 'life_stats' in `LifeStatsObservation` class.
I'll modify my code and update soon.

Miffyli:
👍 Would it also be possible to modify the .diff file manually to add your changes (like done [here](https://github.com/minerllabs/minerl/pull/638/files#diff-b79b398905391ba4a128743fbd0af4b04f05a70381c5b68cf20ccb952d7e8287))? You can also share the modified chunks here and I can do that manual change :)

Miffyli:
Just a small heads-up: currently busy with work, but this PR is not forgotten! I do have things on TODO list, but it may take till the weekend 😅 . Same for the other PR you made.

PR #665, opened 2022-09-03
Add equipped_item observation handler on Java side

xieleo5:
I tried to use the `equipped_item` observation and found that it is not working. And I found that the java side code haven't been implemented yet. So I just tried to add these functions on java side. 

image

The diff file is generated by `generate_mcp_patch.sh` script. There seems to be lots of order diffrence in that file so it seems to be changed a lot. 

Also I know that just adding filter to `getAgentHandler` seems not a clean way. But I really need that observation space so I just add these code to make it work.


Miffyli:
Thanks for this PR too, need to check deeper into the diff :). Looks like we should come up with better way to prepare the diff file because it will continue to break like this in future... And yup, as you have noticed, many of the things that used to be supported are not in v1.0 because it is completely new code on Java side, and Python side has lots of remnants of things that do not work. Will try to look into this at some point but might be too big thing to handle soon (other deadlines are pressing).

xieleo5:
@Miffyli I had directly add change the diff file in the new commit~

Miffyli:
Hey! Sorry for taking _this_ long; I first waited for our competition to finish but then got distracted by the afterwork 😅 . LGTM and thank you for your contributions, again!

PR #712, opened 2023-06-01
Replace np.bool with bool (makes np > 1.20 usable)

JD-The-65th:
Self explanatory.

(Note: This was apparently the only instance of np.bool that search found, not sure if there's more. This is also untested atm.)


Miffyli (approved):
Yup that seems to be the only one. Thanks! :)

PR #756, opened 2024-03-19
Update gradle-wrapper.properties

lewiswlker:


Miffyli:
Hey, happy to try this out. Could you provide bit more context as to why this is required?

lewiswlker:
I am so sorry.I am a new in github.I was not familiar with these in git and this was my wrong operation.thank you for your consideration.You can ignore my mistake:)

PR #763, opened 2024-04-16
Update setup.py to handle the situation when $GRADLE_USER_HOME is not "~"

Aurora-1412:
When $GRADLE_USER_HOME is not "~", the `asset_dir` in setup.py wiil be a wrong path. 

The PR adapts to the situation when ".gradle" directory is not under "~" directory (User customized environment variables $GRADLE_USER_HOME)


Miffyli (approved):
LGTM! Thanks for adding this support (I am not an avid gradle user, so things like this easily slip through 🙃 )

PR #773, opened 2024-06-09
Update setup.py

Aurora-1412:
Sorry for not commenting `asset_dir = os.path.join(os.path.expanduser('~'), '.gradle', 'caches', 'forge_gradle', 'assets')`, hich will make changes(support $GRADELE_USER_HOME) to setup.py ineffective.


Miffyli (approved):
LGTM! That's I get for not carefully reading the code (nor testing it) 🙃 . 

Made a small change because you forked off older code, but now it simply deletes the extra line

PR #775, opened 2024-06-12
Fix warning and update version

GF-Huang:
Fix warning.
Update version.


Miffyli:
Thanks for the updates! Can you confirm what are the minimum Python versions that support these changes? I like the idea of supporting newer things, but given the age of the library, we want to maintain some backwards compatibility.

GF-Huang:
Only tested on Python 3.11.9 and works fine.

jxiong21029:
`logging.warn()` has been deprecated since Python 3.3, see [here](https://bugs.python.org/issue13235).

camelCase aliases for threading (e.g. `setDaemon`) still exist to maintain compatibility with Python 2 code and have been deprecated since Python 3.10, see [here](https://bugs.python.org/issue43723).

